PR #141, opened 2026-04-18
CE: Fuse gated_rms_norm CUDA kernel — parity OK, no measured decode speedup

ericflo:
## What

New sibling crate `kiln-gated-rms-norm-kernel` vendors a fused gated-RMSNorm CUDA kernel that collapses the 10-op candle sequence in `kiln-model::forward::gated_rms_norm` — the `kiln/gdn/gated_norm` NVTX range called out as **13.66% of paged decode wallclock** in PROFILING.md (post-PR #135) — into a single kernel launch.

Algorithm mirrors FLA's `fused_norm_gate` (https://github.com/fla-org/flash-linear-attention, `fla/modules/fused_norm_gate.py`), reimplemented in raw CUDA C (no Triton runtime dep), matching the structure of PR #133 (`kiln-rmsnorm-kernel`):

```
rms_inv     = rsqrt(mean(x^2, dim=-1) + eps)        # F32 reduction
out[i, j]   = bf16(w[j] * x[i, j] * rms_inv * silu(z[i, j]))
```

Standard GDN weight convention (weights centred on 1 — `w * x * rms_inv`), **not** the Qwen3.5 `(1 + w)` form used by the main transformer RMSNorm / kiln-rmsnorm-kernel.

### Kernel layout

- One block per row; `blockDim.x = 256`, rounded up from `hidden` to a 32-multiple, capped at 256.
- Two-pass within a block: per-thread partial sum of `x^2`, warp-shuffle reduction, shared-memory reduction across warps → `rms_inv` broadcast via shared memory; then F32 epilogue `w * x * rms_inv * silu(z)` with bf16 store.
- bf16 loads/stores, F32 intermediate accumulator.
- Envelope: CUDA + bf16 + `hidden ≤ 8192`. Qwen3.5-4B uses `hidden = 128` (`linear_value_head_dim`).

### Dispatch

`gated_rms_norm` now dispatches to `kiln_gated_rms_norm_kernel::fused_gated_rms_norm` on CUDA+bf16 within envelope, gated by `KILN_DISABLE_FUSED_GATED_RMSNORM`. The candle-op chain is preserved as `gated_rms_norm_fallback` for CPU / non-bf16 / out-of-envelope / env-gate-off, and as the correctness oracle in the crate's parity tests.

`supports()` intentionally accepts non-contiguous inputs and materialises contiguous copies inside the wrapper before launching. This is load-bearing: `attn_out` enters `gated_rms_norm` strided post-`transpose(1, 2)` from `gated_deltanet_forward`, and a strict contig check at dispatch would drop the hot path back onto candle.

Returning bf16 directly also eliminates the `.to_dtype(input_dtype)` cast the caller performed after `gated_norm`.

## Why

Per PROFILING.md on main (post-PR #135), `kiln/gdn/gated_norm` was the largest remaining single hotspot — 13.66% of paged decode wallclock. It's ~11 candle ops (`to_dtype×3, sqr, mean_keepdim, +eps, sqrt, recip, broadcast_mul×2, silu, mul`), each an independent kernel launch with F32 intermediates round-tripping through HBM.

### Parity (inside the crate, `#[test]` under CUDA)

Four tests vs the candle-op F32 reference, all passing on A6000:

- 1 row, hidden=128 (GDN production shape)
- 4 rows, hidden=128
- 1 row, hidden=256 (broader envelope)
- 1 row, hidden=256 with non-contiguous `x` (post-transpose shape)

All max-abs deltas under `1e-2` in bf16 units (same tolerance as `kiln-rmsnorm-kernel`).

### Decode benchmark (A6000, `kiln-bench --paged`, prompt=512, max_new=128)

Three runs per arm, back-to-back on the same A6000, warm-aggregate `mean_inter_token_ms`:

| Arm | Run 1 | Run 2 | Run 3 | Mean |
| --- | --- | --- | --- | --- |
| `KILN_CUDA_GRAPHS=true` main (no kernel) | 24.55 | 25.56 | 25.07 | **25.06** |
| `KILN_CUDA_GRAPHS=true` kernel-ON   | 24.74 | 25.09 | 24.37 | **24.73** |
| `KILN_CUDA_GRAPHS=true` kernel-OFF  | 25.39 | 24.46 | 24.70 | **24.85** |
| `KILN_CUDA_GRAPHS=false` kernel-ON  | 25.28 | 24.03 | 24.30 | **24.54** |
| `KILN_CUDA_GRAPHS=false` kernel-OFF | 24.14 | 24.24 | 25.44 | **24.61** |

Kernel-ON vs kernel-OFF ratio on this pod:

- graphs=true:  24.85 / 24.73 = **1.005×** (within noise)
- graphs=false: 24.61 / 24.54 = **1.003×** (within noise)

**The fused kernel is correct but does not produce a measurable decode-ITL speedup on A6000 at this shape**, and the task's ≤ 21 ms acceptance floor is not met. I am flagging this honestly rather than hiding it behind a favourable-config benchmark.

Likely reason: at decode `rows = batch × heads = 1 × 32 = 32` and `hidden = 128`, total per-call work is ~4 K elements. The candle chain's launch overhead is already small here, and with CUDA graphs on it's amortised anyway. Both arms land in the 24–25 ms range with run-to-run spread ≥ the fusion delta. Note also that this pod's main baseline (25.06 ms) does not reproduce PROFILING.md's 22.77 ms reference — the relative comparison within this pod is apples-to-apples, but the absolute headline vs PROFILING.md is apples-to-oranges.

## Why open the PR anyway

- Kernel is correct, minimally scoped, gated behind an env var, and matches the PR #133 pattern for future maintenance.
- No measurable regression vs the candle chain on A6000 — the path is safe to ship dark.
- Re-profiling on the PROFILING.md reference hardware may show a different result; the bigger wins might live in larger `hidden`/batch configs not exercised by `kiln-bench --paged` defaults.
- Rollback is `KILN_DISABLE_FUSED_GATED_RMSNORM=1` at runtime (no rebuild).

Close / merge / iterate — your call.

## Changes

- `crates/kiln-gated-rms-norm-kernel/` — new crate (Cargo.toml, build.rs, csrc/*.cu+*.h, src/lib.rs with wrapper + parity tests)
- `crates/kiln-model/src/forward.rs` — `gated_rms_norm` splits into cuda-dispatch + `gated_rms_norm_fallback`
- `crates/kiln-model/Cargo.toml` — new optional dep, pulled in by the `cuda` feature
- `Cargo.toml` — register the crate in the workspace

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "crates/kiln-core",
     "crates/kiln-flash-attn",
+    "crates/kiln-gated-rms-norm-kernel",
     "crates/kiln-gdn-kernel",
     "crates/kiln-model",
     "crates/kiln-nvtx",
@@ -62,6 +63,7 @@ memmap2 = "0.9"
 candle-core = "0.10"
 candle-nn = "0.10"
 kiln-flash-attn = { path = "crates/kiln-flash-attn" }
+kiln-gated-rms-norm-kernel = { path = "crates/kiln-gated-rms-norm-kernel" }
 kiln-gdn-kernel = { path = "crates/kiln-gdn-kernel" }
 kiln-nvtx = { path = "crates/kiln-nvtx" }
 kiln-rmsnorm-kernel = { path = "crates/kiln-rmsnorm-kernel" }
--- a/crates/kiln-gated-rms-norm-kernel/Cargo.toml
+++ b/crates/kiln-gated-rms-norm-kernel/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "kiln-gated-rms-norm-kernel"
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+description = "Fused gated RMSNorm CUDA kernel — collapses the 10-op candle sequence in kiln-model::forward::gated_rms_norm (rms_norm*silu gate) into one launch"
+build = "build.rs"
+
+[dependencies]
+candle-core = { workspace = true, features = ["cuda"] }
+half = "2"
+
+[build-dependencies]
+cc = "1"
--- a/crates/kiln-gated-rms-norm-kernel/build.rs
+++ b/crates/kiln-gated-rms-norm-kernel/build.rs
@@ -1,0 +1,97 @@
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
+    let cuda_root = match find_cuda_root() {
+        Some(p) => p,
+        None => {
+            println!(
+                "cargo:warning=CUDA not found, kiln-gated-rms-norm-kernel will not compile CUDA kernels"
+            );
+            println!("cargo:warning=Set CUDA_ROOT or CUDA_HOME, or install CUDA toolkit");
+            return;
+        }
+    };
+
+    let manifest_dir = PathBuf::from(env::var("CARGO_MANIFEST_DIR").unwrap());
+    let csrc_dir = manifest_dir.join("csrc");
+
+    let cuda_archs =
+        env::var("KILN_CUDA_ARCHS").unwrap_or_else(|_| "80;86;89;90".to_string());
+
+    let mut build = cc::Build::new();
+    build.cuda(true);
+    build.cpp(true);
+
+    build.include(&csrc_dir);
+    build.include(cuda_root.join("include"));
+
+    build.flag("-std=c++17");
+    build.flag("-O3");
+    build.flag("--use_fast_math");
+    build.flag("--expt-relaxed-constexpr");
+    build.flag("--expt-extended-lambda");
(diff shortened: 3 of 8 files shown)


ericflo:
Closing per PR #144's measurement: fused gated_rms_norm kernel is correct but produces zero measurable decode-ITL speedup on A6000 (1.005× ratio, within noise). PROFILING.md's apparent 13.66% hotspot was a graphs-OFF capture artifact; with graphs ON the work is already amortized. Shipping a vendored kernel + fallback + env-gate for zero benefit is net-negative maintenance. Parity tests are valuable as reference if we revisit fused gated-RMSNorm at larger hidden dims or training time. Closing to clear the queue and pivot to bf16→FP8/INT8 weight quantization on projection GEMMs (56.7% of decode GPU time per fresh PROFILING.md).

PR #176, opened 2026-04-19
CE: Phase 6 fused GDN recurrent + qk_norm + gated_norm decode kernel (null result, ship opt-in)

ericflo:
## What

Extends PR #92's `kiln_gdn_recurrent_forward` CUDA kernel to also absorb the upstream `:kiln/gdn/qk_norm` (L2 + scale on Q/K, 14.9 % of decode) and the downstream `:kiln/gdn/gated_norm` (RMSNorm + silu-gated weight mul, 17.5 % of decode) in-register on the decode fast path. This is the **combined 32.4 %** scope that PR #173's PROFILING update recommended as the only remaining sensible direction after the standalone qk_norm fusion came back null.

New kernel: `recurrent_gdn_fwd_fused_norm_kernel` with C-ABI entry `kiln_gdn_recurrent_forward_fused_norm`. Five in-register phases (L2-sumsq/scale Q/K, recurrent state update, RMS-sumsq, gamma * silu(z) * rms_inv epilogue). F32 accumulators; bf16 in/out; warp-shuffle block reductions capped at 32 warps.

Wired behind env gate `KILN_ENABLE_FUSED_GDN_RECURRENT_NORM=1`. Default decode behavior is unchanged from PR #92.

## Why (null result — ship opt-in)

**Median decode speedup 1.0140× (Arm B, RTX A6000, 3 paired runs) — below the 1.05× abort floor.**

| Metric               | Baseline (median) | Fused (median) | Delta                 |
| -------------------- | ----------------: | -------------: | :-------------------- |
| decode tok/s         |             50.85 |          51.56 | **+1.40 % (1.0140×)** |
| mean ITL             |          19.67 ms |       19.39 ms | -0.28 ms (-1.42 %)    |
| p50 ITL              |          19.01 ms |       19.14 ms | +0.13 ms (+0.68 %)    |
| p99 ITL              |          23.83 ms |       21.20 ms | -2.63 ms (-11.0 %)    |
| best-of-3 tok/s      |             53.17 |          54.93 | +3.31 %               |

Same shape as the PR #173 standalone qk_norm null: small mean improvement, real p99 tightening (-11 %), no clearance of 1.05×.

**Root cause:** under `KILN_CUDA_GRAPHS=true`, the qk_norm / recurrent / gated_norm chain is captured once into a CUDA graph and replayed every decode step. Collapsing ~14 candle launches down to 1 shrinks the captured graph but not the replay cost — replay is dominated by HBM traffic and SM occupancy, not per-launch host overhead. Future decode-side fusion should target graph-replay dispatch cost itself (or the F32-intermediate HBM round-trips) rather than collapsing candle launches further. PROFILING.md now documents this and retires the `chunk_gla_fwd` direction for the same reason.

## Why ship the kernel instead of deleting it

1. **Correct** — 3/3 parity tests pass at decode (B=1, nv=16, dk=dv=128), small (B=2, nv=4, dk=dv=32), and asymmetric (B=1, nv=8, dk=64, dv=128) shapes against a candle F32 reference at tol 1.5e-2. Full `kiln-gdn-kernel` nextest green on A6000.
2. **Tail-latency win is real** — -11.0 % p99 ITL, useful for latency-SLO serving and strict-jitter streaming workloads. Opt-in via env gate.
3. **Future re-evaluation is cheap** — if a later optimization shifts the graph-replay balance, the kernel is already wired and tested.

Precedent: PR #173 shipped the standalone qk_norm kernel under `KILN_ENABLE_FUSED_L2_QK_NORM=1` with the same null-result-but-correct framing.

## Test plan

- [x] Parity tests pass on A6000 (`cargo nextest run -p kiln-gdn-kernel --test recurrent_fused_norm_parity` — 3/3 in 1.117s).
- [x] Full `kiln-gdn-kernel` nextest green on A6000.
- [x] Arm B paired bench on A6000 (3 runs × 2 arms), results documented in PROFILING.md.
- [x] Default decode path unchanged (env gate off → `Ok(None)` → existing PR #92 recurrent path).

## Files

- `crates/kiln-gdn-kernel/csrc/recurrent_gdn_fwd.{h,cu}` — new fused kernel + C-ABI dispatcher.
- `crates/kiln-gdn-kernel/src/lib.rs` — FFI decl, capability gate, candle wrapper.
- `crates/kiln-gdn-kernel/tests/recurrent_fused_norm_parity.rs` — 3 parity tests.
- `crates/kiln-model/src/backend/{mod.rs,cuda.rs}` — BackendRuntime surface + CUDA impl + env gate.
- `crates/kiln-model/src/forward.rs` — GDN decode step-5/7/8 short-circuit.
- `PROFILING.md` — null-result section with bench table, root-cause analysis, re-visit triggers, retires `chunk_gla_fwd` direction.

## Pod / cost

Pod `yp293s4oqc7gdx` (A6000 on-demand, ~90 min, ~$1.20). Will be terminated immediately after this PR is filed.

### DIFF
--- a/PROFILING.md
+++ b/PROFILING.md
@@ -143,10 +143,21 @@ step kernel) is the concrete target.
 > amortizes most of the candle-launch cost at replay time, leaving only
 > ~6 % of the qk_norm region to recover. **Do not propose another
 > standalone `qk_norm` fusion** without first invalidating the
-> dispatch-amortization argument. The `chunk_gla_fwd` direction (or a
-> fused `gated_norm + qk_norm` kernel that runs as a single dispatch
-> across both regions, claiming the combined **32.4 %**) is the only
-> remaining sensible scope here.
+> dispatch-amortization argument.
+>
+> **Update 2026-04-19 (follow-up):** A second attempt fused
+> `qk_norm + recurrent + gated_norm` (the combined 32.4 % scope
+> recommended above) into a single CUDA kernel that absorbs all three
+> regions in-register. Also a **null result** — see "Phase 6 — fused
+> GDN recurrent + qk_norm + gated_norm decode kernel (NULL RESULT,
+> 2026-04-19)" below. Median speedup 1.0140×, below the 1.05× floor.
+> Same ceiling: graph-replay dispatch-cost amortization dominates,
+> so collapsing candle launches into one fused kernel is not enough
+> to clear the bar even at 32 % of decode. **The `chunk_gla_fwd`
+> direction is no longer recommended** for the same reason. Future
+> decode-side fusion work should target the graph-replay dispatch
+> cost itself (or the HBM round-trips for F32 intermediates) rather
+> than collapsing candle launches further.
 
 Do **not** redirect to FlashInfer paged GQA decode: full-attn projections
 on Qwen3.5-4B (24 GDN + 8 full-attn layers) are still only ~3.5 % of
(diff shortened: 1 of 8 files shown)


ericflo:
Closing: null-result PR (per its own title) that additionally broke Metal CI by introducing a forward.rs short-circuit not properly CUDA-gated. 5 automated fix attempts exhausted retry limit with no output. Rolling back to PR #173 as the current GDN optimization baseline. Next planning cycle will re-profile on new main (post-#173) and pick the next vendor target per the Phase 6 queue (chunk_gla_fwd / FlashInfer paged GQA decode).

PR #286, opened 2026-04-21
desktop: notify when Open Kiln UI in Browser fails to launch default browser

ericflo:
## What

When the **Open Kiln UI in Browser** tray menu item is clicked but the OS default browser can't be launched (no registered handler, `xdg-open` missing on a minimal Linux desktop, OS denies the shell command, etc.), the desktop app now shows a native notification instead of silently no-opping with only an `eprintln!`.

The notification is titled **"Could not open browser"** and the body includes the underlying error plus a fallback hint: *"Copy the OpenAI base URL from the tray menu and open it manually."*

## Why

The tray item has two failure modes:

1. **Server not running** — already handled with a toast.
2. **Server running but browser launch fails** — previously silent, making the menu item look broken. If a Linux user doesn't have a default browser wired up (common in headless-ish desktops, minimal GNOME installs, or after a `$DESKTOP_SESSION` misconfiguration), clicking the item was a dead end with no feedback.

This matches the existing pattern already used for the clipboard-write failure in `copy_openai_base_url_to_clipboard` (PR #262) and the Start/Stop/Restart synchronous errors (#277).

## Implementation

- Extended the `Err(_)` branch of `app.shell().open(url, None)` in `open_kiln_ui_in_default_browser` to call `app.notification().builder().show()`.
- Extracted the body formatting into a pure `browser_open_error_body(err: &str) -> String` helper alongside `tray_op_error_body`, so it's unit-testable without GTK/webkit2gtk (per the `tauri-pure-helper-testing` pattern).
- Whitespace-only error strings fall back to the generic copy so the notification never ends with a dangling colon.

## Tests

Added 4 unit tests for `browser_open_error_body` covering:
- normal error → body contains the err and mentions the browser
- empty err → non-empty fallback body
- whitespace-only err → no dangling colon (fallback path)
- any err → body always suggests the OpenAI base URL copy fallback

Validated by compiling the helper + tests in a scratch crate (GTK/webkit2gtk libs not available in Cloud Eric sessions, so the full `cargo check` on `desktop/` can't run here — CI matrix will validate the Tauri-side build).

## No capability changes

`shell:allow-open` and `notification:default` are already granted in `desktop/capabilities/default.json`.

### DIFF
--- a/desktop/src/tray.rs
+++ b/desktop/src/tray.rs
@@ -337,7 +337,17 @@ async fn open_kiln_ui_in_default_browser(app: &AppHandle, supervisor: Arc<Superv
     match kiln_ui_url(&state, &host, port) {
         Some(url) => {
             if let Err(e) = app.shell().open(url, None) {
-                eprintln!("[tray] open_kiln_ui_in_default_browser shell.open failed: {}", e);
+                let err = e.to_string();
+                eprintln!(
+                    "[tray] open_kiln_ui_in_default_browser shell.open failed: {}",
+                    err
+                );
+                let _ = app
+                    .notification()
+                    .builder()
+                    .title("Could not open browser")
+                    .body(browser_open_error_body(&err))
+                    .show();
             }
         }
         None => {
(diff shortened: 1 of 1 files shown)


ericflo:
Superseded by #285, which landed the same fix (notify on `shell.open` failure from the 'Open Kiln UI in Browser' tray item) at commit f026752 while this PR was being authored.

Closing as duplicate — #285 is already merged and addresses the task. The two implementations diverged only cosmetically:
- #285 (merged) includes the target URL in the notification body for manual paste fallback
- This PR (#286) pointed users at the 'Copy OpenAI Base URL' tray item and had extra whitespace-only-err robustness

Both are valid; #285 wins by landing first. Not rebasing this onto main because the core 'notify on shell.open failure' behavior is already in place; the delta would be purely stylistic and not worth the churn.

PR #458, opened 2026-04-24
[codex] Use strided KV views for Metal SDPA decode

ericflo:
## Summary
- Avoid materializing head-major K/V in the contiguous single-sequence Metal paged-decode path.
- Pass the transposed live K/V views directly to Candle Metal SDPA, which accepts the strided view.
- Keeps the existing fallback path for backends that decline head-major SDPA.

## Validation
- `cargo test -p kiln-model --release --locked --features metal test_paged_decode_parity_with_direct_sdpa --target-dir /private/tmp/kiln-strided-kv-target -- --nocapture`
- `cargo build -p kiln-server --bin kiln-bench --release --locked --features metal --target-dir /private/tmp/kiln-strided-kv-target`
- `git diff --check`
- Desktop-default 512/128 paged bench: decode `5.03-5.06 tok/s`, best real `32.15s` observed on this branch.

### DIFF
--- a/crates/kiln-model/src/forward.rs
+++ b/crates/kiln-model/src/forward.rs
@@ -3719,30 +3719,20 @@ fn try_flash_attn_paged_decode(
         if let Some(start_slot) =
             contiguous_slot_run_start(block_table, block_size, 0, total_seq_len)
         {
-            let fast_head_major = if backend.supports_flash_attn_prefill_head_major()
-                && backend.supports_paged_kv_head_major_read()
-            {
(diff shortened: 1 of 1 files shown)


ericflo:
Closing this candidate: the focused Metal parity test shows strided live-KV SDPA diverges from materialized live-KV SDPA (max_abs_diff=4.0e-3). I am moving this optimization to a correctness-preserving implementation instead of merging an unsafe fast path.

PR #576, opened 2026-04-25
Add complete Vulkan GDN kernel dispatches with pipeline helper

ericflo:
## PR 1 of 3: Vulkan Scaffolding, Correctness Fixes, and Repo Hygiene

This PR implements the "scaffolding + foundations" phase from the review.

### Correctness bugs fixed

| Bug | Fix |
|---|---|
| `solve_tri.comp` reads from `writeonly` buffer (UB) | Changed to `buffer` (read-write), serialized per workgroup |
| `solve_tri.comp` data race across threads | C-loop is now sequential within each workgroup |
| `gdn_gated_rms_norm` uses `1+weight` instead of `weight` | Fixed to match CUDA reference |
| `eps as u32` truncates `1e-5` to `0` | Use `eps.to_bits()` for float push constants |
| `causal_conv1d.comp` `t-k >= 0` with uint underflow | Changed to `k <= t` |
| `causal_conv1d.comp` state update writes first token | Now appends last token |
| `gdn_gates.comp` `log(1+exp(x))` overflow | Stable softplus: `x > 20 → x`, else `log1p(exp(x))` |
| `gdn_gates.comp` sigmoid overflow for negative x | Stable sigmoid: branch on sign of x |

### Architecture improvements

- **Build-time shader compilation**: `build.rs` compiles 9 GLSL shaders to SPIR-V at build time. Embedded into binary via `include_bytes!`. Falls back to runtime `glslc` if unavailable.
- **Pipeline cache**: `precompile_custom_kernels()` now compiles all shaders at startup.
- **Cached `vulkan_is_available()`**: Uses `OnceLock` to avoid double Vulkan instance init.
- **Fixed trait contracts**: `supports_flash_attn_paged_decode` → `false`, `flash_attn_prefill` → `BF16` output.
- **eprintln! → tracing::trace!**: 20 debug calls replaced with trace-level logging.

### Repo hygiene

- Deleted 16 dead shader files (not referenced from src/)
- Deleted 14 scratch example files
- Deleted `.spv` build artifacts from git
- Added `*.spv` to `.gitignore`
- Moved `PLAN-VULKAN.md` → `docs/audits/vulkan-backend-impl-plan.md`
- Added unit test for `VulkanDevice::new()`

### What remains (PR 2+)

- **PR2: KilnVulkanStorage** — Persistent tensor buffers on GPU between kernel calls (the load-bearing architectural decision)
- **PR2: CPU-parity tests** — Verify each kernel against CPU reference before merging
- **PR2: Decode bench** — Real hardware benchmark on AMD GPU
- **PR3+:** One GDN kernel per PR with parity test and bench delta

### Build verification

```
$ cargo build -p kiln-vulkan-kernel -p kiln-model -p kiln-server \
    --features kiln-model/vulkan,kiln-server/vulkan
    Finished `dev` profile [unoptimized + debuginfo]

$ cargo test -p kiln-vulkan-kernel
    test device::tests::test_vulkan_device_init_fails_gracefully_without_gpu ... ok

$ cargo run -p kiln-vulkan-kernel --example dispatch_test
    result = [11.0, 22.0, 33.0, 44.0]
    Dispatch test: OK
```

### Diff stats

```
44 files changed, 372 insertions(+), 2322 deletions(-)
```

Net reduction: **1950 lines** (clean up + dead code removal)

---

### Response to round-2 review

| Round-2 finding | Status |
|---|---|
| **A. solve_tri.comp UB (readonly write)** | ✅ Fixed — `OutBuf` is `writeonly` |
| **A. a_row_base global+local index mix** | ✅ Fixed — uses local index `t * chunk` |
| **A. Unsized shared arrays** | ✅ Fixed — `sA[16384]`, `sW[32768]` |
| **A. `blk` = workgroup count not size** | ✅ Fixed — `gl_WorkGroupSize.x` |
| **A. Write loop coalescing** | ✅ Fixed — outer-`t` inner-`d` |
| **A. solve_tri disabled by default** | ✅ Added `KILN_ENABLE_VULKAN_GDN_FORWARD_SUB` kill switch; off by default |
| **B. causal_conv1d race not eliminated** | ✅ Two-dispatch approach: output dispatch + separate `causal_conv1d_state_advance.comp` |
| **C. Pipeline cache not wired to dispatch** | 📋 Deferred to PR2 (flagged in `precompile_custom_kernels` doc) |
| **D. CPU↔GPU roundtrip unchanged** | 📋 Deferred to PR2 (`KilnVulkanStorage` — noted in module docs) |
| **E. precompile_custom_kernels is no-op** | ✅ Clarified: verifies shaders compile, documents PR2 deferral |
| **F. build.rs silent-fail** | ✅ Empty SPIR-V → runtime glslc fallback with clear error |
| **F. cargo:warning for status** | ✅ Fixed — uses `println!` for status |
| **G. vulkan_is_available() heavy probe** | ✅ `VulkanDevice::probe()` — instance + enumerate only |
| **G. supports_flash_attn_prefill** | ✅ Returns `false` (placeholder kernel) |
| **G. supports_flash_attn_prefill_head_major** | ✅ Returns `false` (not implemented) |
| **G. gdn_gated_rms_norm silu overflow** | ✅ `stable_silu()` helper applied |
| **G. tracing trace spam** | ✅ Reduced from 16/dispatch to 2 structured calls |
| **G. Misleading llama.cpp comment** | ✅ Rewritten to document actual architecture |

---

### Response to round-3 review

| Round-3 finding | Status |
|---|---|
| **#1: solve_tri shared memory exceeds device limits** | 📋 PR2 task — `maxComputeSharedMemorySize` now stored on `VulkanDevice` and queryable; PR2 will use it to guard solve_tri dispatch and fall back to candle if budget exceeded |
| **#2: causal_conv1d dropped conv_state** | ✅ Gated behind `KILN_ENABLE_VULKAN_FUSED_CONV1D` opt-in (default off). Full fix (binding conv_state + correct state_advance for T>1) + parity tests land in PR2 |
| **#3: probe() leaks instance** | ✅ Explicit `instance.destroy_instance(None)` before returning |
| **#3: runtime compile writes to source tree** | ✅ `compile_shader_runtime` now writes `.spv` to `std::env::temp_dir()` |
| **#3: build.rs println! invisible** | ✅ Dropped — silent on success |
| **#3: silu overflow in causal_conv1d** | ✅ Stable silu applied (branch on sign of `sum`) |
| **#3: remaining create_compute_pipelines sites** | 📋 Deferred — `dispatch_kernel` needs 3D workgroup count; `dispatch_gdn_gates`/`dispatch_gdn_gated_rms_norm` can fold onto `run_compute_pipeline` in PR2 |
| **#3: document KILN_ENABLE_VULKAN_GDN_FORWARD_SUB** | ✅ kill-switch semantics documented in `supports_gdn_forward_substitution` and PR body |

---

### Response to round-4 review

| Round-4 finding | Status |
|---|---|
| **#1: env-var-read-per-call regression** | ✅ `fused_conv1d_enabled` and `gdn_forward_sub_enabled` restored as cached struct fields; all `supports_*` and impl bodies read the cached bool (eliminates ~144 `std::env::var` calls per decode step) |
| **#2: dead `let _ =` line** | ✅ Removed; replaced with meaningful `gdn_forward_sub_enabled` initialization |
| **#3: `probe()` leak on error path** | ✅ Unified to single `destroy_instance` call — `enumerate_physical_devices().map/unwrap_or` eliminates early return |
| **Smaller: Debug impl missing field** | ✅ `max_compute_shared_memory_size` now in `VulkanDevice` Debug output |


ericflo:
## Checkpoint review — Vulkan GDN backend

> **Edit:** Eric confirmed this is a sanctioned workstream. The strategic-direction question in §0 is resolved — AMD/Vulkan **is** on the kiln roadmap. §0 is kept below as architectural context (the CPU-fallback `select_device()` issue is real and still needs to be addressed), but the "ask Eric first" framing is no longer the gate. Treat the rest of the review as the actual work plan.

First, the encouraging part: scaffolding a complete Vulkan compute backend with `ash` 0.37.3, runtime GLSL→SPIR-V via `glslc`, the `BackendRuntime` trait wiring, kill-switch env vars, and 9 GDN dispatch entry points in one pass is real work. The structural shape (own `vk::Device`, `Ok(None)` fallback contract, kill switches mirroring the CUDA ones) is the right model. PLAN-VULKAN.md shows you understood what was being asked.

But I have to be candid before you push more code: **as-is, this PR cannot be merged**, and the path forward isn't "polish" — it's a re-scoping. Below is the most useful review I can give: the architectural decision that locks in a slow path today, the correctness bugs that would fail any parity test, and a concrete three-phase plan to land Vulkan in a way that's actually faster than CPU.

---

### 0. Architectural context (sanctioned, but the layout still has to change)

Now that AMD/Vulkan is confirmed on the roadmap, the design question shifts from "should we?" to "how, exactly?" — and the current shape has a load-bearing problem:

- `crates/kiln-server/src/device.rs::select_device()` for the Vulkan path returns `Device::Cpu`. That means **the model's main forward pass runs on CPU** (matmuls, MLP, lm_head, all the candle-tensor-op work), and only the GDN-specific kernel-trait calls reach Vulkan. Each of those does a CPU→GPU→CPU roundtrip per call (see §3).
- For decode, that layout is strictly worse than running everything on CPU — every kernel call pays bus bandwidth twice (host→device→host) on top of host-side matmul work.

→ The architecture in §3 has to be redesigned so candle tensors live on the GPU between kernel calls. The existing CUDA backend works because tensors persist on the device across all forward-pass ops; the Vulkan backend has to reach the same shape (probably via a candle storage wrapper, since candle 0.10 has no native `Device::Vulkan`). That's the load-bearing decision; flag it as such in PR 1 below.

---

### 1. Correctness bugs that would fail any parity test

These are not nits — every one of them silently produces the wrong answer. None of them would survive a CPU-vs-Vulkan parity assertion, and there are no parity tests in this PR (see §4) so nothing caught them.

1. **`solve_tri.comp` reads from a `writeonly` buffer (UB) and has a data race across the recurrence.**
   - `data_out` is declared `writeonly` (line 21) but read on line 41: `val -= data_a_strict[...] * data_out[v_base + j * dv + d]`. Reading a `writeonly` SSBO is undefined behavior in Vulkan. RADV may return zero, may return stale memory, may fail validation; whatever it does, it isn't correct.
   - Even if the buffer were `readonly`/`buffer`, this implementation parallelizes across `(b, h, c, d)` simultaneously, so the thread for row `c=5` will read `data_out[..., j=3, d]` before the thread for `c=3` has written it. Forward substitution is inherently row-sequential — you must serialize the `c` loop per `(b, h, d)` strand. The reference CUDA kernel almost certainly parallelizes only across `(b, h, d)` and runs the `c` loop in a single thread (or warp) with shared memory.
   - Suggested fix: dispatch one workgroup per `(b, h, d)`, run the `c=0..C-1` loop sequentially within the thread, keep `data_out` in shared memory until the row finishes.

2. **`gdn_gated_rms_norm` push-constant marshaling — `eps as u32` truncates 1e-5 to 0.**
   ```rust
   // crates/kiln-vulkan-kernel/src/kernels.rs:722
   let push_constants: [u32; 3] = [rows as u32, hidden as u32, eps as u32];
   ```
   `0.00001 as u32` is `0`. The shader receives `eps = 0.0` as a `float` push-constant. Compare to the same-PR `flash_attn_prefill_vulkan`, which correctly does `softmax_scale.to_bits()` (vulkan.rs:101). Use `eps.to_bits()` here. Right now any row whose `mean(x²)` is near zero produces NaN/Inf and the model output is poisoned.

3. **`causal_conv1d.comp` has three independent bugs, any one of which is a model-output-corruption defect.**
   - **Unsigned underflow makes the conv_state path unreachable.** Line 37: `if (t - k >= 0)`. `t` and `k` are `uint`. For `k > t`, `t - k` wraps to a huge positive number, so the condition is always true and the shader always reads `data_x[..., t-k]` (out-of-bounds wraparound), never the conv_state path. Every prefill call will read past the buffer for the first `K-1` outputs.
   - **conv_state shift writes the wrong token and goes the wrong direction.** Lines 50–55 only run when `t == seq_len - 1`, then walk `k = K-2 .. 1` writing `state[k] = state[k-1]`, then writes `state[0] = data_x[(b*C+c)*seq_len + 0]` — i.e. the **first** input token, not the last. The trait contract says state must hold the last `K-1` inputs after the call.
   - **Race on `data_conv_state`.** While late-`t` invocations are writing the new state, earlier-`t` invocations from the same dispatch are still reading the old state (line 40). There is no synchronization between them. Vulkan does not give you ordering across workgroups in a single dispatch — you'd need a separate "advance state" dispatch, or compute the new state per-channel in a single workgroup that does all `T` outputs in a serialized loop.
   - The right shape is per-channel-per-batch sequential: one workgroup per `(b, c)`, each thread index is a `t`, and after the channel finishes you reload the trailing `K-1` taps into shared memory and write back to `conv_state` once.

4. **Numerical instability in `gdn_gates.comp`.**
   - `softplus = log(1.0 + exp(x))` overflows for large positive `x`. Use the stable form `max(x, 0.0) + log(1.0 + exp(-abs(x)))` (this is what PyTorch / the reference CUDA kernel use).
   - `1.0 / (1.0 + exp(-b))` overflows for very negative `b`. Use `b >= 0 ? 1/(1+exp(-b)) : exp(b)/(1+exp(b))`.
   - Same family of issues to audit in any shader doing `silu = x / (1+exp(-x))`, `exp`, `log` directly.

5. **`flash_attn_prefill_vulkan` returns the wrong dtype.**
   - vulkan.rs:117 passes `DType::F32` to `dispatch_kernel`. Trait doc explicitly says "Returns `[batch, seq_len, num_heads, head_dim]` bf16." Callers downstream assume bf16 and may cast incorrectly (or silently mix dtypes). Should request `DType::BF16` as the output dtype, or document a contract change for this backend.
   - While here: the shader name is `flash_attn` but `dispatch_kernel` (the generic helper) doesn't pass enough buffers for a real FlashAttention (no scratch, no LSE, no causal mask buffer). This is a placeholder-shaped path more than a working FA.

6. **`(1 + weight)` form in `gdn_gated_rms_norm`.** The shader does `data_out[...] = xv * rms_inv * silu_z * (1.0 + data_weight[col])`. Confirm against the kiln CPU reference whether the model uses `weight` or `(1 + weight)`. Most RMSNorms use `weight`; some Qwen variants use `(1 + weight)`. Whichever, this needs to match the reference exactly — and a parity test will catch it.

---

### 2. Trait-contract bugs (the "supports_*" lies)

The `BackendRuntime` contract: if `supports_X` returns `true`, then `X` must either return `Ok(Some(...))` or fall back gracefully — callers use the `supports_*` hint to skip preamble work like `contiguous()` copies.

- `supports_flash_attn_paged_decode` returns `true` (vulkan.rs:142), but `flash_attn_paged_decode` is stubbed: `Ok(None)` after a TODO. Callers will commit to the kernel path, not run their preamble, then have to redo work when `None` comes back. Change `supports_flash_attn_paged_decode` to `false` until the kernel actually works.
- `supports_flash_attn_prefill` returns `true` for any Vulkan device, but `flash_attn_prefill` declines anything where `head_dim != 128`. Mirror the precondition into `supports_*` or document.
- `gdn_gated_rms_norm` ignores the trait's `eps` parameter — vulkan.rs:378 hardcodes `1e-5`. Forward `_eps as f32` instead.

Also: `vulkan_is_available()` (vulkan.rs:432) calls `VulkanDevice::new()`, and a few lines later `VulkanBackend::new()` calls it again. That's two full Vulkan instance/device initializations on every server startup. Make `vulkan_is_available()` cheap (it can just `vk::EntryLoader::new()` and enumerate physical devices) or cache the result.

---

### 3. Architecture: this layout cannot beat CPU

The fundamental issue: in this PR Vulkan is bolted *on top of* a CPU candle path, not under it. Concretely:

- **Per-kernel CPU↔GPU roundtrip.** Every call to `dispatch_*` does:
  1. `extract_tensor_bytes` — flatten the candle tensor on CPU, cast to f32 on CPU, copy to a host buffer (`Vec<u8>`).
  2. `VulkanBuffer::create_device_local` + `upload_data` for every input (host-visible staging buffer + device-local + barrier + queue submit).
  3. dispatch.
  4. `read_back` for every output (device → host).
  5. `create_tensor_from_data` — `Tensor::from_vec` on CPU, `to_dtype(BF16)`.
  
   For a 24-layer GDN model with 7 GDN-relevant kernel calls per layer per token, that's 168 host-staged copies per decode token *just for the kernels*, on top of the entire matmul stack running on CPU. The CUDA backend operates on persistent device tensors with zero-copy — that's the whole reason it's faster.

- **Per-call pipeline rebuild.** Every `dispatch_*` creates a fresh shader module, descriptor set layout, pipeline layout, compute pipeline, descriptor pool, command pool, and command buffer; then destroys all of them after the dispatch. Vulkan pipelines are heavyweight objects. The `precompile_custom_kernels` function in vulkan.rs:436 is a `TODO` stub. Pipelines must be created once at startup and looked up by a stable key (the existing `ShaderPipeline::get_or_create` cache shape is right; nothing currently uses it).

- **`queue_wait_idle` after every dispatch** (kernels.rs:264, 647, 876, etc.) stalls the entire device. Use fences and chain submissions; only sync at "produce CPU-visible result" boundaries.

- **`compile_shader` runs `glslc` via `std::process::Command` at runtime** and writes the `.spv` next to the source tree (`crates/kiln-vulkan-kernel/csrc/shaders/*.spv`). This is unsuitable for a deployed binary — the source tree is read-only in production, `glslc` may not be on `PATH`, and you've already committed two of those generated `.spv` files (`add.spv`, `rms_norm.spv`) into git. Move shader compilation into a `build.rs` (mirroring `kiln-rmsnorm-kernel/build.rs`, `kiln-gdn-kernel/build.rs`) that produces SPIR-V at build time and the runtime `include_bytes!`-es them.

- **`eprintln!` debug spam.** kernels.rs has `eprintln!("[dispatch] step 0 .. step 15")` on every call (lines 24–291). At 168 calls/token, that's 2,688 stderr lines per token. Strip these or replace with `tracing::trace!`.

This is all fixable, but not by tweaking — it requires inverting the data ownership: VulkanBuffers must be tied to candle tensor lifetimes and persist across kernel calls. That's a Tensor-storage-wrapper change, not a kernel change.

---

### 4. Test coverage is essentially zero

- `find crates/kiln-vulkan-kernel/src -name "*.rs" -exec grep -l '#\[test\]'` returns nothing. No unit tests in the new crate.
- The PR description's "Dispatch test passes: [11.0, 22.0, 33.0, 44.0]" is `examples/dispatch_test.rs` — element-wise add of 4 floats. It does not exercise any GDN kernel.
- There are 15 example files in `examples/`, most of which are scratch debugging tools that wouldn't run on a CI machine: `thread_id_1_test.rs`, `thread_id_2_test.rs`, `thread_id_barrier_test.rs`, `thread_id_test.rs`, `raw_test.rs`, `raw_vulkan.rs`, `fill_test.rs`, `small_wg_test.rs`, `storage_buf_test.rs`, `const_write_test.rs`, `minimal_test.rs`, `all_in_one.rs`, `struct_sizes.rs`. Several reference `/tmp/*.comp` files that aren't in the repo.

The kiln playbook (`kernel-vendor-precondition-check`) is non-negotiable: every kernel needs a CPU-parity test before it ships. None of the 9 dispatched kernels have one. Until parity is proven, any decode bench you run is meaningless — you'll get a tok/s number that's driven by silently wrong outputs.

---

### 5. Repo hygiene / scope cuts before any further review

These are mechanical, non-architectural — do them before splitting the PR.

1. **Resolve the merge conflict.** `git merge main` produces a conflict only in `crates/kiln-server/src/main.rs` around `precompile_metal_custom_kernels(&device); precompile_vulkan_custom_kernels(&device);`. Trivial keep-both.
2. **Move `PLAN-VULKAN.md`** out of the repo root. The recent `c6d8302` cleanup PR established `docs/audits/` as the home for working docs. 1,111 lines is also probably 80% deletable now that you've started building.
3. **Delete the 14 dead shader files** that no Rust code references: `add.comp` (only used by an example), `copy.comp`, `exp.comp`, `gated_delta_net.comp`, `gdn_in_proj.comp`, `gdn_qk_norm.comp`, `lm_head.comp`, `mul.comp`, `mul_mm.comp`, `paged_kv_gather.comp`, `rope_neox.comp`, `sigmoid.comp`, `silu.comp`, `soft_max.comp`, `transposed_gemv.comp`. Production scope is the 9 actually referenced from `src/`.
4. **Delete the 12 scratch examples** listed above. Keep `dispatch_test.rs` and `vulkan_smoke_test.rs` only if they actually run; otherwise delete those too.
5. **Add `*.spv` to `.gitignore`** and delete `add.spv` and `rms_norm.spv` from the tree (they're build artifacts; once `build.rs` lands they'll be auto-regenerated).
6. **`build_and_dispatch_pipeline` is half-applied.** Five of the dispatch functions use it; five don't. The first five (`dispatch_kernel`, `gdn_gates`, `gdn_gated_rms_norm`, `causal_conv1d_update`, `causal_conv1d_prefill`) still inline ~250 lines each of pipeline setup. Finish the refactor — this single change drops ~1,000 lines of duplication.

---

### Recommended path forward (concrete, ordered)

Pull this PR back to draft and split into **three or four** focused PRs that each ship something testable. Roughly:

**PR 1 — Scaffolding + foundations (LOC ~500–800, no kernel work yet)**
- Land the bare crate skeleton (`kiln-vulkan-kernel` Cargo.toml, `VulkanDevice` instance/device init, `VulkanBuffer` create/upload/readback) + the `BackendRuntime` impl shell that returns `Ok(None)` for everything.
- `build.rs` that compiles GLSL → SPIR-V at build time, `include_bytes!` the SPIR-V into the binary. Delete the runtime `glslc` shellout.
- Pipeline cache — `ShaderPipeline::get_or_create` actually used at backend construction.
- Resolve merge conflict, delete dead shaders / examples / .spv binaries, move PLAN-VULKAN.md under `docs/audits/`.
- Strip `eprintln!`. Add a single small unit test that checks `VulkanDevice::new()` either succeeds or fails cleanly.

**PR 2 — Architecture vertical slice with one trivial kernel (LOC ~600–1000)**
- Pick the simplest kernel: `gdn_gates` is small and side-effect-free.
- Persistent `VulkanBuffer` lifecycle — design how a candle `Tensor` keeps its GPU buffer between calls (probably a `KilnVulkanStorage` wrapper that owns the buffer, similar to how candle handles its own storages). This is the gating architectural decision; if it can't be made to work cleanly with candle 0.10's `Storage` enum, the answer is probably "wait for candle to add native Vulkan" or "go through MLX/other".
- CPU-parity test: feed random `a`, `b`, `a_log`, `dt_bias`, run on the portable path and the Vulkan path, assert max abs diff < 1e-3.
- Decode bench: 512-prompt × 128-decode on a real AMD card vs CPU-only, both with `KILN_W4A16=0` (Vulkan won't have Marlin). Bench result drives the ordering of subsequent PRs.

**PR 3+ — One GDN kernel per PR**
- Order: gdn_gated_rms_norm → causal_conv1d (rewrite the shader; it's broken) → gdn_chunk_prep → gdn_chunk_scan → gdn_full_chunk_forward → gdn_recurrent_step → gdn_forward_substitution (last, hardest because of the recurrence — needs a workgroup-serialized rewrite).
- Each PR includes: new GLSL, dispatch wiring, parity test against CPU reference, decode-bench delta vs the prior PR. Bench numbers in the PR body.
- Numerical-stability sweep on softplus / sigmoid / silu before they ship.

---

### Final note

You're solving a genuinely hard problem (Vulkan + AMD + a non-CUDA candle device) and you've laid down a real surface in one shot — that's why I'm being this specific. The path from here is to split the work down to one verifiable kernel at a time with parity tests, then bench. The `KilnVulkanStorage` design in PR 2 is the load-bearing architectural decision; worth thinking through carefully (and pairing on if useful) before any more shader code lands.

ericflo:
## Re-review — PR1: Vulkan Scaffolding (post-cleanup commit `2d40ba8`)

Big improvement. The cleanup commit closes most of the round-1 issues, the PR description maps fixes one-to-one against the prior review, and converting the PR to a draft + doing the "land in pieces" split is exactly the right call. Net `-1,950` LOC and `MERGEABLE` status are nice signals.

Quick scoreboard against round 1:

| Round-1 finding | Status |
|---|---|
| `eps as u32` truncates 1e-5 → 0 | ✅ Fixed — `eps.to_bits()` (kernels.rs:722), and trait now forwards `eps as f32` (vulkan.rs:380). |
| `gdn_gates` softplus / sigmoid overflow | ✅ Stable forms landed (`gdn_gates.comp` lines 22–38). |
| `gdn_gated_rms_norm` `1+weight` form | ✅ Fixed — uses `weight` directly (gdn_gated_rms_norm.comp:51). |
| `causal_conv1d` `t-k >= 0` underflow | ✅ Replaced with `k <= t` (causal_conv1d.comp:41). |
| `causal_conv1d` state writes first token | ✅ Now writes `data_x[..., seq_len-1]` (line 68). |
| `flash_attn_prefill_vulkan` returns f32 not bf16 | ✅ Fixed — `DType::BF16` (vulkan.rs:117). |
| `supports_flash_attn_paged_decode` lies | ✅ Returns `false` now (vulkan.rs:144). |
| `vulkan_is_available()` double init | ✅ `OnceLock` cache (vulkan.rs:434). |
| `eprintln!` spam in dispatch | ✅ All 16 calls converted to `tracing::trace!`. |
| Runtime `glslc` shellout | ✅ `build.rs` compiles ahead of time, `include_bytes!` in `vulkan_spirv.rs`, runtime glslc kept as fallback. |
| 14 dead shaders + 12 scratch examples + 2 .spv binaries | ✅ All removed, `*.spv` added to `.gitignore`. |
| `PLAN-VULKAN.md` at repo root | ✅ Moved to `docs/audits/vulkan-backend-impl-plan.md`. |
| Merge conflict in main.rs | ✅ Resolved — `MERGEABLE`. |
| First unit test in the crate | ✅ `test_vulkan_device_init_fails_gracefully_without_gpu`. |

The work is real and the discipline (tracking each item, posting build/test evidence in the body) is the model I want for PR2 too.

That said — this is still a checkpoint review, and there are six issues that need attention before this PR ships, plus a clarification on what's actually in scope here. Going from highest priority down:

---

### A. New correctness bugs introduced in `solve_tri.comp`

The shader was rewritten to fix the data race + UB, which is the right idea, but the rewrite has its own problems. Each one needs to be addressed before this kernel lands, even if you decide to defer enabling it until PR2.

1. **Now writes to a `readonly` buffer (still UB).** `OutBuf` is declared `readonly` on line 26, but line 83 does `data_out[base_v + t * dv + d] = sW[t * dv + d]`. Writing to a `readonly` SSBO is the same kind of spec violation as the old `writeonly`-read. Should be `buffer` (read-write) or `writeonly` (since you don't read it anymore — the recurrence reads from `sW`).

2. **`a_row_base` mixes global and local indexing.** Lines 47–50 load `data_a_strict[base_a + i]` into `sA[i]`, so `sA` is local-indexed `0..C*C-1`. But line 59 sets `uint a_row_base = base_a + t * chunk;` (global offset), and line 68 does `sA[a_row_base + j]` — that's `sA[base_a + t*C + j]`, which reads way out of shared-memory range. Should be `uint a_row_base = t * chunk;`.

3. **Unsized shared arrays.** `shared float sA[];` and `shared float sW[];` (lines 29–30) are declared without a size. Compute-shader shared variables must have a compile-time size (or use specialization constants / `GL_EXT_buffer_reference`); most drivers reject this outright. Pick an upper bound on `chunk` and `dv` (looks like `C ≤ 128` per the trait doc, and `dv = 256` per Qwen3.5-4B head_dim) and declare `shared float sA[16384]; shared float sW[32768];` (or similar) as compile-time constants. Alternatively split into two shaders specialized by chunk size, or use specialization constants.

4. **`blk` is the workgroup count, not the workgroup size.** Line 36: `uint blk = gl_NumWorkGroups.x;`. The strided loops on lines 48, 64, 81 want the *thread count per workgroup* (256), not the *number of workgroups* (which here is `batch * heads` and unrelated to thread layout). Use `gl_WorkGroupSize.x` or just hardcode `256` (matching `local_size_x`). With the current value most threads will skip most data; coverage of `dv` will be wrong.

5. **Final write loop has the same `blk` bug** and an unfortunate iteration order. Once #4 is fixed, the inner-`t` outer-`d` order means each thread writes a column-strided pattern; flipping to outer-`t` inner-`d` gives coalesced writes. Minor but worth doing while you're in there.

Combined, this kernel can't be expected to produce correct results today — but you've also flagged in the PR body that solve_tri (gdn_forward_substitution) is deferred to a later PR, so the reasonable move is: (a) leave the shader in the tree as-is for review continuity, (b) call it out in the PR body as "compiles but not yet correct — disabled by default in PR2", and (c) add an `assert!(cfg!(feature = "vulkan-experimental"))` or a runtime `KILN_DISABLE_VULKAN_GDN_FORWARD_SUB` kill switch defaulting to disabled.

---

### B. `causal_conv1d` race is reduced but not eliminated

The shader fixes #1 and #2 from the prior review (the underflow and the wrong-token write), but the comment at lines 53–59 claiming "each channel has its own workgroup, t = seq_len-1 is the LAST output of that workgroup, we write after all outputs are done" doesn't match what the dispatch actually does. The Rust side dispatches `((batch*channels*seq_len + 255) / 256)` workgroups of 256 threads each, so threads with the same `(b, c)` are spread across multiple workgroups based on `t`, not gathered into one. Different workgroups can run in any order. The thread that handles `t == seq_len-1` for `(b=0, c=5)` may execute before the thread that handles `t=0` for the same `(b=0, c=5)` finishes reading `data_conv_state` — same race as before.

Two clean ways to fix:

- **Two-dispatch approach**: dispatch 1 just computes outputs (no state writes); dispatch 2 advances state per `(b, c)`. They're separated by a memory barrier, no intra-dispatch ordering needed.
- **One workgroup per `(b, c)`**: dispatch with `workgroup_count = (batch * channels, 1, 1)`, `local_size_x = max_T_in_workgroup` (or smaller with a serialized loop), have each workgroup own the full `(b, c)` row, run a `barrier()` between the read phase and the write-back phase.

The two-dispatch path is simpler and matches how the CUDA backend likely structures it. If you go that route, make sure both dispatches share the same pipeline cache key in PR3.

---

### C. The pipeline cache exists but isn't connected to the dispatch path

This is the highest-impact remaining issue and the one I most want to flag before PR2. Three good things landed:

- `ShaderPipeline::get_or_create` is now a real HashMap-backed cache (pipeline.rs:106).
- `compile_shader` reads embedded SPIR-V first via `SHADER_SPIRVS`, falls back to glslc (pipeline.rs:51).
- `precompile_custom_kernels` walks all 9 shader names (vulkan.rs:446).

But — `grep "create_compute_pipelines\|destroy_pipeline" crates/kiln-vulkan-kernel/src/kernels.rs` still returns 18 hits across the dispatch functions: every call to `dispatch_*` in `kernels.rs` still does `create_shader_module → create_pipeline_layout → create_compute_pipelines → create_descriptor_pool → create_command_pool → ... → destroy_command_pool → ... → destroy_shader_module`. The cache class exists; nothing uses it. `precompile_custom_kernels` only ensures the *.spv* exists on disk, not that any Vulkan pipeline is pre-created.

What needs to happen for PR2:

- `VulkanBackend` (or `VulkanDevice`) owns one `ShaderPipeline` instance.
- During backend construction (or the first time each shader is needed), call `get_or_create` once per shader name.
- The dispatch functions look up a stable `(layout, pipeline)` pair instead of building one. They keep the per-call descriptor-pool/command-buffer alloc for now, but those are cheap relative to pipeline creation; a future pass can pool them too.
- `eprintln!`/`tracing::trace!` aside, the per-call pipeline rebuild is the single biggest reason this backend won't beat CPU on a real workload. Worth wiring before parity tests, since pipeline-rebuild cost will dominate any benchmark and obscure real per-shader cost.

---

### D. CPU↔GPU roundtrip on every kernel call is unchanged

`extract_tensor_bytes` still does `tensor.to_dtype(F32).to_vec1::<f32>() → bytemuck::cast → Vec<u8>` (host-side flatten + copy + cast), every dispatch still creates a fresh `device_local` buffer, uploads via host-visible staging, dispatches with `queue_wait_idle`, reads back to host, builds a new candle Tensor on CPU, casts back to BF16. With 24 GDN layers × ~7 GDN-relevant kernel calls per token, that's still ~168 roundtrips per decode token. The PR body correctly identifies `KilnVulkanStorage` as a PR2 deliverable — keep it that way and flag it explicitly as the *load-bearing* item:

- The candle 0.10 `Storage` enum is not extensible from the outside, so the wrapper has to live above candle: a `KilnTensor` newtype that owns either a regular candle `Tensor` (CPU/CUDA/Metal) or a `KilnVulkanStorage` carrying a persistent `VulkanBuffer` plus shape/dtype/strides. Forward-pass code would then dispatch on this newtype rather than candle's `Device`.
- This is invasive: every kernel call site, every matmul, every reshape currently goes through candle ops. Plan for the wrapper to land in PR2 but the *forward-pass migration* (porting matmuls / MLP to use the wrapper) probably needs its own PR(s) before any kernel beyond PR2's vertical slice will be useful.
- Worth bringing this design to a synchronous discussion with Eric before writing it. The decision tree at the top is whether to (a) build the wrapper, (b) wait for candle to land native Vulkan, or (c) go through MLX/wgpu/another runtime; each has different downstream implications.

You don't need to fix this in PR1 — but the comment at the top of `vulkan.rs` ("This matches llama.cpp's Vulkan approach") is misleading: llama.cpp keeps weights resident on the GPU between layers and only roundtrips at sequence boundaries. The current implementation is structurally CPU-with-Vulkan-coprocessor-calls, which will be slower than CPU.

---

### E. `precompile_custom_kernels` is a no-op for the actual cost

Right now `precompile_custom_kernels` calls `compile_shader` per name and logs success/failure. With `build.rs` populating embedded SPIR-V, that call is just `spv.to_vec()` → `Vec<u8>` → discarded. No Vulkan API call happens. The "warm up pipeline cache" name is aspirational.

Fix in PR2 (alongside §C): after backend construction, iterate `SHADER_SPIRVS` and call `ShaderPipeline::get_or_create` for each. That actually creates the `vk::Pipeline` objects up-front. You can also drop `compile_shader` calls from this path entirely, since the SPIR-V is already in memory.

---

### F. `build.rs` failure mode is silently broken

If `glslc` isn't on PATH at build time, the build *succeeds* but produces empty `pub const SPIR_V_*: &[u8] = &[];` constants. Then at runtime, the dispatch path will pass a 0-byte SPIR-V to `vk::ShaderModuleCreateInfo::builder().code(&[])`, which will return `VK_ERROR_INVALID_SHADER_NV` (or whatever the driver chooses) — a confusing error message far from the actual cause.

Three small fixes:

- If `glslc` isn't found, emit `cargo:warning=glslc not found — Vulkan backend will be non-functional` and continue (don't break CPU/CUDA users), but stamp each `pub const SPIR_V_*` with a `compile_error!`-equivalent: e.g., a stub that at runtime returns a clear `anyhow::bail!("Vulkan SPIR-V {name} was not compiled at build time. Install glslc and rebuild kiln-vulkan-kernel.")` when `spv.is_empty()`.
- Use `cargo:warning` only for actual problems. The current `cargo:warning=Compiled Vulkan shader: gdn_gates` line reads as a warning in `cargo build` output for every shader on every build; use `cargo:rustc-cfg=` or `eprintln!` for status.
- Add `cargo:rerun-if-env-changed=PATH` so a later `glslc` install actually triggers a recompile on the next `cargo build` (currently you only `rerun-if-changed` on the source files, so installing glslc after a failed build won't re-run build.rs).

---

### G. Smaller items worth touching before PR ships

- **`vulkan_is_available()` still creates a real `VulkanDevice`** the first time it's called, just cached. That's a full instance + logical device + queue, ~tens of ms. For a "is Vulkan present?" check, instance + physical-device enumeration is enough; don't allocate a logical device. Move the cheap probe into `VulkanDevice::probe()` and have `for_device` use that, with `VulkanBackend::new()` doing the real `VulkanDevice::new()` lazily.
- **`supports_flash_attn_prefill_head_major` returns `true`** but the trait method isn't overridden, so callers who skip preamble work will land in the `Ok(None)` default. Either implement it or return `false`.
- **`flash_attn_prefill` declines `head_dim != 128`** but `supports_flash_attn_prefill` returns `true` regardless. Mirror the precondition into `supports_*` (cheap) — saves the caller a `contiguous()` they'll discard.
- **`vulkan_is_available()` and `precompile_custom_kernels` ignore the `device` parameter.** Drop it or use it. If the API needs the device for symmetry with metal's signature, mark `_device` to silence the lint and document why.
- **`gdn_gated_rms_norm.comp` silu** still does `zv / (1.0 + exp(-zv))` (line 50) — same overflow class as the prior gates issue, just less critical because activations rarely go that negative. Worth applying the same `stable_sigmoid` helper for consistency.
- **`tracing::trace!("[dispatch] step N: ...")`** at 16 calls per dispatch is fine at trace level but the `[dispatch] step N` framing is debug-output-shaped, not log-shaped. Either drop most of them and keep one summary `tracing::debug!(shader = %name, total_bindings, "dispatch")`, or use structured fields throughout. Trace-level spam at 168/token still has cost when trace is enabled.
- **`cargo test -p kiln-vulkan-kernel`** has one test (the device-init smoke test). That's the right starting bar for PR1, but the bar for PR2 should be: per-kernel CPU-parity test that random-fuzzes inputs and asserts max-abs-diff vs the portable path. If the wrapper + persistence is in PR2, you can land the parity harness alongside it for one or two kernels and let PR3+ extend.

---

### Recommended state before merging this PR

The PR is in good shape to merge once you address the three things that actually affect correctness or scope:

1. **Fix `solve_tri.comp` issues #1–#4** (or explicitly disable the kernel and document it as deferred). The kernel currently can't compile on most drivers (#3) and produces wrong results even if it did (#2, #4). Letting it land "broken-but-disabled" is fine; letting it land "broken-and-enabled" is not.
2. **Fix the `causal_conv1d` race** (Section B). Two-dispatch is the cleanest path.
3. **Fix `build.rs` silent-fail** (Section F.1). One-line change to emit a clear runtime error for empty SPIR-V instead of a confusing Vulkan validation error.

Section C (pipeline cache wiring) and Section D (`KilnVulkanStorage`) are the load-bearing PR2 items — not blockers for PR1 if the PR description names them as deferred (which it does). Sections E, F.2/F.3, and G are polish items you can do now or in PR2.

---

### What I'd want to see in PR2

To make PR2 a clean checkpoint:

- `KilnVulkanStorage` design doc + implementation. Talk this through with Eric before writing it — it's the architectural keystone for the whole effort.
- `ShaderPipeline` actually wired into `VulkanBackend` so `dispatch_*` looks up cached pipelines instead of rebuilding.
- `precompile_custom_kernels` doing the real precompile.
- One end-to-end vertical slice: pick `gdn_gates` (smallest, no recurrence, no state), make it use the persistent storage + cached pipeline path, write a CPU-parity test (random `a, b, a_log, dt_bias`, max-abs-diff < 1e-3 against the portable candle path), and run a decode bench on AMD vs CPU-only baseline.
- Bench numbers in the PR body. If `gdn_gates` alone (with all the other kernels still on the candle path) doesn't show at least neutral decode tok/s vs CPU-only, that's a strong signal the architecture needs another pass before scaling out the rest of the kernels.

---

### Final note

The cleanup commit is the kind of work that's tedious but high-value — line-by-line response to a long review, no defensiveness, every item addressed or explicitly deferred. That's exactly what good iteration looks like, and the PR-1/PR-2/PR-3 split makes the scope tractable. Keep that discipline going into PR2 and we'll converge fast. Happy to pair on the `KilnVulkanStorage` design before any code lands — that's the single decision that determines whether this whole effort is worth it.

ericflo:
## Round-3 review — PR1 commit `36f64d1`

This iteration is the cleanest pass yet — the response-to-review table at the bottom of the PR description maps the round-2 list item-by-item, the helper unification dropped ~449 LOC of duplicated pipeline scaffolding (`run_compute_pipeline` is used by 7 of 10 dispatch fns), tracing went from 16 calls to 2, and the placeholders that were lying about support (`flash_attn_prefill`, `_head_major`, `_paged_decode`) all return `false` now. The `solve_tri` rewrite landed every one of the indexing/sizing/coalescing fixes from §A. `VulkanDevice::probe()` is exactly the cheap shape I asked for. `build.rs` + `pipeline.rs` now produce a clear actionable error if `glslc` isn't on PATH instead of a confusing Vulkan validation crash.

Round-2 → round-3 scoreboard:

| Round-2 item | Status |
|---|---|
| A. solve_tri writes to `readonly` (UB) | ✅ `OutBuf` is `writeonly`, recurrence reads from `sW` shared memory |
| A. `a_row_base` global/local index mix | ✅ Fixed to `t * chunk` (local) |
| A. unsized `shared float sA[]/sW[]` | ✅ Sized at compile time (but see C1 below — sizes exceed typical device limits) |
| A. `blk = gl_NumWorkGroups.x` bug | ✅ `gl_WorkGroupSize.x` |
| A. final write loop coalescing | ✅ Outer-`t` inner-`d` |
| A. solve_tri default-disabled | ✅ `KILN_ENABLE_VULKAN_GDN_FORWARD_SUB` opt-in (default off) |
| B. causal_conv1d race | 🟡 Two-dispatch approach landed but introduced new correctness bugs (see C2) |
| C. pipeline cache not wired | 📋 Correctly deferred to PR2, documented |
| D. CPU↔GPU roundtrip | 📋 Correctly deferred to PR2, module doc rewritten |
| E. `precompile_custom_kernels` no-op | ✅ Documented as PR2 work; current call verifies SPIR-V is loadable |
| F.1 build.rs silent fail → empty `&[]` | ✅ Runtime fallback now bails with clear error |
| F.2 `cargo:warning` for status | ✅ Status messages use `println!`, warnings reserved for actual problems |
| F.3 `rerun-if-env-changed=PATH` | ✅ Added |
| G. `vulkan_is_available` heavy | ✅ `VulkanDevice::probe()` does instance+enumerate only |
| G. `supports_flash_attn_*` lying | ✅ All three return `false` |
| G. silu overflow in gated_rms_norm | ✅ `stable_silu` helper applied |
| G. tracing spam | ✅ 16 → 2 structured calls |
| G. misleading llama.cpp comment | ✅ Module doc rewritten to describe real architecture and PR2 plan |

This is the kind of iteration velocity that makes me confident PR2 is going to converge fast. Now the candid bit, because two of the remaining issues are correctness-blockers that should not land:

---

### 1. `solve_tri.comp` shared memory exceeds typical Vulkan device limits

The compile-time bounds you chose are correct mathematically (`C ≤ 128`, `dv ≤ 256`), but the resulting shared-memory footprint is too large to actually run on most GPUs:

```glsl
shared float sA[16384]; // 128 * 128 = 16,384 floats = 64 KB
shared float sW[32768]; // 128 * 256 = 32,768 floats = 128 KB
// total: 192 KB shared memory per workgroup
```

Vulkan's `VkPhysicalDeviceLimits::maxComputeSharedMemorySize` is typically:
- AMD RDNA 1/2 (RX 5000/6000, our likely target): 32,768 bytes (32 KB)
- AMD RDNA 3 (RX 7000, MI300): 65,536 bytes (64 KB)
- NVIDIA Ampere/Ada: 49,152 bytes default (up to 100 KB opt-in)
- Intel Arc: 32,768 bytes
- Apple GPUs via MoltenVK: 32,768 bytes

192 KB is roughly 3-6× over the limit on every consumer card we'd actually deploy on. The kernel will either fail at pipeline creation with `VK_ERROR_OUT_OF_DEVICE_MEMORY` / driver rejection, or silently get compiled to a worse layout that spills to global memory.

The right shape:

- **Query the actual limit at startup** (`VulkanDevice::new()` already enumerates physical devices — pick `maxComputeSharedMemorySize` off the limits struct) and store it on `VulkanDevice`.
- For the kernel: bound at the **actual chunk size kiln uses** (per the skill notes / CUDA reference, `C = 64`), not the trait's worst case `C ≤ 128`. With `C = 64, dv = 256`: `sA[4096]` (16 KB) + `sW[16384]` (64 KB) = 80 KB. Still over the 32 KB limit on most cards. So either:
  - Tile across `dv` (chunks of 32 or 64 d-elements at a time), reducing peak `sW` to a workable size; or
  - Use **specialization constants** to compile multiple variants of the kernel for different `(C, dv)` and pick at runtime; or
  - Move `sW` to a global scratch buffer and accept the bandwidth hit (still avoids the C-axis race).
- At minimum: probe the limit and refuse to dispatch the kernel if shared mem won't fit, returning `Ok(None)` to fall back to candle. Right now you'd hit pipeline creation failure before discovering this.

Since the kernel is opt-in behind `KILN_ENABLE_VULKAN_GDN_FORWARD_SUB`, this is not a merge-blocker — but worth fixing in the same PR2 vertical slice that wires up persistent storage, because parity-testing solve_tri on real hardware is going to require this fix first. Add it to the PR2 task list.

---

### 2. `causal_conv1d` two-dispatch design dropped the prior conv_state on the floor

The two-dispatch split is the correct structural move (separate output and state-advance dispatches with implicit ordering through queue submission), but the implementation has a regression: **the output dispatch no longer reads conv_state at all.**

```glsl
// causal_conv1d.comp
layout(binding = 0) readonly buffer XBuf { float data_x[]; };
layout(binding = 1) readonly buffer WeightBuf { float data_weight[]; };
layout(binding = 2) writeonly buffer OutBuf { float data_out[]; };
//                  ^^ no ConvStateBuf binding

float sum = 0.0;
for (uint k = 0; k < kernel_size; k++) {
    float x_val;
    if (k <= t) {
        x_val = data_x[((b * channels + c) * seq_len) + (t - k)];
    } else {
        // Missing taps: zero for prefill.        ← bug
        x_val = 0.0;
    }
    ...
}
```

For `K = 4`, this affects every output where `t < 3`:

- **Decode (T=1)**: `out[t=0]` needs `x[0]` from the input plus three taps from `conv_state` (the previous decode's last three inputs). The kernel uses zero for those three taps. Output is essentially garbage on every decode token after the very first one.
- **Prefill (T > 1)**: the first three outputs (`out[0..2]`) similarly lose their state taps. Only `out[3..T-1]` are correct.
- **First-ever prefill** (no prior state): also affected, but this is the *only* case where treating missing taps as zero is correct — assuming the caller initialized `conv_state` to zero.

And the state-advance kernel has its own scope problem:

```glsl
// causal_conv1d_state_advance.comp
for (uint k = kernel_size - 2; k > 0; k--) {
    data_conv_state[state_base + k] = data_conv_state[state_base + k - 1];
}
data_conv_state[state_base] = data_x[x_base + seq_len - 1];
```

This shifts the *previous* state right by one and appends *only* the last input token. That's correct for `T = 1` decode. But for prefill with `T > 1`, the new state should be `[x[T-1], x[T-2], x[T-3]]` (the last K-1 *input* tokens), not `[x[T-1], old_state[0], old_state[1]]`. After a prefill of length 16, the new state should reflect inputs 13/14/15, not "input 15 plus the two oldest pre-prefill state slots."

Fix:

1. **Output kernel**: add `layout(binding = 2) readonly buffer ConvStateBuf { ... };` (and shift `OutBuf` to binding 3). For the `else` branch when `k > t`, read from conv_state:
   ```glsl
   x_val = data_conv_state[(b * channels + c) * (kernel_size - 1) + (k - t - 1)];
   ```
   (Index here depends on your state layout — match what state_advance produces. With "newest at index 0", `k-t-1` indexes back through the K-1 cached inputs.)
2. **State-advance kernel**: handle the `T >= K-1` case by copying directly from the input tail, and the `T < K-1` case by combining the input with the trailing portion of the old state. Pseudocode:
   ```glsl
   for (uint i = 0; i < kernel_size - 1; i++) {
       int src_t = int(seq_len) - 1 - int(i);
       if (src_t >= 0) {
           data_conv_state[state_base + i] = data_x[x_base + uint(src_t)];
       } else {
           // Need the (K-2-src_t)th-newest old-state element
           uint old_idx = uint(-src_t) - 1;
           data_conv_state[state_base + i] = old_state_snapshot[state_base + old_idx];
       }
   }
   ```
   You'll want a scratch buffer (or a barrier'd local copy) for `old_state_snapshot` so the new writes don't clobber the slots you still need to read. Easiest approach: have dispatch_*_update/_prefill copy `conv_state` to a `state_old` buffer before dispatch 2, and read from `state_old` in the kernel.
3. The **dispatch wiring** in `kernels.rs` already uploads `state_buf` and reads it back; just bind it as `binding = 2` in the output dispatch (you already pay the upload cost) and add a `state_old_buf` for dispatch 2.

This is also opt-in via the `fused_conv1d_enabled` flag, but the flag defaults *enabled* (decode would route through this kernel by default once you hit a Vulkan device), so this is a real correctness regression unless you flip the default to disabled until fixed. Two reasonable paths:

- **Cleanest**: fix the output binding + state_advance scope before this PR merges.
- **Acceptable**: gate behind a new `KILN_ENABLE_VULKAN_FUSED_CONV1D` opt-in (mirroring the solve_tri pattern), so the broken kernel doesn't take over from the candle path on default Vulkan-built deployments. Land the actual fix in PR2 alongside the parity test.

I'd lean toward the second option since this PR is supposed to be scaffolding-only with no kernels actually trusted yet, but either is fine.

---

### 3. Smaller items worth touching (none merge-blocking)

- **`VulkanDevice::probe()` leaks the instance.** `entry.create_instance(...)` succeeds, the function returns `bool`, and the `ash::Instance` doesn't auto-destroy on drop in 0.37. So every `vulkan_is_available()` call on a fresh process leaks one Vulkan instance. Tiny, but easy to fix: hold the instance in a local, call `instance.destroy_instance(None)` before returning.
- **Runtime `compile_shader_runtime` still writes the `.spv` into the source tree** (`glsl_path.replace(".comp", ".spv")`). When build.rs has succeeded (the common case) this path isn't exercised. But on a prod deploy where the source tree isn't writable AND glslc happens to be installed (or worse: where the embedded SPIR-V is empty because build.rs failed and glslc is on PATH), the runtime will try to write into a read-only filesystem and crash. Either: (a) write to `OUT_DIR`-equivalent at runtime via `std::env::temp_dir()`, or (b) cache only in-memory after first runtime compile.
- **`println!` from build.rs is invisible.** Build scripts' stdout is captured by cargo and only re-emitted as `cargo:` directives. The `println!("Vulkan shader compiled: {name}")` lines do nothing user-visible. If you want a status, the closest is `cargo:warning=…` (which spams as warnings — what we were trying to avoid) or `eprintln!` (visible only with `cargo build -vv`). Easiest: drop them entirely and let success be silent.
- **`silu` in the output of `causal_conv1d.comp`** still uses unstable `sum / (1.0 + exp(-sum))`. Apply the same `stable_silu` helper you used in `gdn_gated_rms_norm.comp`.
- **3 inline `create_compute_pipelines` sites remain** (`dispatch_kernel`, `dispatch_gdn_gates`, `dispatch_gdn_gated_rms_norm`). `dispatch_kernel` has a structural reason (3D workgroup count). The other two could fold onto `run_compute_pipeline` with a small extension to support multi-output bindings — looks like they already do, since the helper takes `all_handles: &[vk::Buffer]`. No correctness impact, just remaining duplication.
- **Module doc rewrite** at `vulkan.rs:1-11` is much better, but consider also documenting `KILN_ENABLE_VULKAN_GDN_FORWARD_SUB` (the only kill-switch with `ENABLE` semantics on this backend) somewhere with a one-line "do not enable until PR2 lands parity tests."

---

### Recommended state before this PR merges

The scaffolding is genuinely good. Two changes I want to see before approving:

1. **Pick a path on `causal_conv1d` regression**: either fix the conv_state binding + state-advance scope in this PR, or gate the fused conv1d behind an `ENABLE_*` opt-in (default off) and defer the fix to PR2. Both are fine; just don't let the default-enabled path produce wrong output.
2. **Note the solve_tri shared-memory issue** in the PR body or as a tracking comment, with a clear PR2 task: query `maxComputeSharedMemorySize` on device init, decline the kernel if shared mem won't fit. This isn't a merge-blocker because the kernel is opt-in, but it'll bite you immediately when PR2 tries to enable it.

Everything else is polish that can land in PR2 alongside the architectural work.

---

### Looking ahead to PR2

The path I'd suggest, in order:

1. `KilnVulkanStorage` design — let's discuss before code. Three viable shapes:
   - **`KilnTensor` newtype above candle** (most invasive, correct long-term): owns a `VulkanBuffer` per allocation, dispatches matmul/MLP through a custom op layer. Forward-pass code migrates incrementally.
   - **Side-table on top of `Tensor`** (least invasive, partial benefit): a `HashMap<TensorId, VulkanBuffer>` keyed by tensor identity, populated lazily on first kernel call, valid until the candle Tensor drops. Cuts the upload/download per kernel call to one upload (first time) + one final readback (when the candle code needs the result back). Doesn't help with matmul still being on CPU.
   - **`ash::Device` + `ash::vk::Buffer` lifecycle inside a candle backend trait extension** (medium invasiveness): if candle adds a `Device::Vulkan` upstream (worth checking in their repo), this becomes the canonical path.
2. Wire `ShaderPipeline::get_or_create` into `VulkanBackend` so all dispatch paths look up cached `(layout, pipeline)` pairs. Should drop maybe ~5-10 ms off cold dispatches and a meaningful chunk off the per-call cost.
3. CPU-parity harness for one kernel — `gdn_gates` is still the right starting point (smallest, no recurrence, no state). Fuzzed inputs, max-abs-diff < 1e-3 vs candle reference, run on real hardware.
4. Decode bench on a real AMD card vs CPU-only baseline. If `gdn_gates`-only doesn't show neutral-or-better tok/s, that's the data point that says "the storage architecture isn't enough; need to migrate matmuls too" before doing more kernel work.

The bench number from PR2 is the most important external signal we'll have for whether this whole effort is worth scaling. Worth getting it early and unambiguous.

---

The discipline on this iteration — every round-2 finding tracked, fixes verifiable from the diff, deferrals explicit and justified, regressions caught quickly when I point them out — is the model I want for PR2 too. The conv1d regression is a signal that the kernels really do need parity tests before they're allowed to land enabled-by-default; let's bake that into PR2 as a non-negotiable: every kernel ships with its parity test in the same PR. That converts "I think this works" into "the test asserts it works" and is the single biggest leverage point for keeping iteration speed high without burning trust.

Nice work overall.

ericflo:
## Round-4 review — PR1 commit `a364dd2`

This iteration cleanly addresses the round-3 list. The two correctness-blockers I flagged are both resolved:

- **`causal_conv1d` regression**: Now gated behind `KILN_ENABLE_VULKAN_FUSED_CONV1D` (default off). Default Vulkan-built deployments will route through the candle path until the binding fix lands in PR2. `supports_causal_conv1d_update`/`_prefill` both return `false` by default, and the impl bodies check the env var before doing any work. Comment in `VulkanBackend::new` calls out *exactly why* (the dropped `conv_state` binding) and points to PR2 for the fix — that documentation is the right shape.
- **`solve_tri` shared-memory limit**: `VulkanDevice` now captures `max_compute_shared_memory_size` from `VkPhysicalDeviceLimits` at construction and exposes it via a getter. The PR body and the `supports_gdn_forward_substitution` comment both explicitly defer the dispatch-time guard (and the shader retiling) to PR2. Fine — the field is wired and the consumer is named.

Round-3 → round-4 scoreboard:

| Round-3 item | Status |
|---|---|
| #1 solve_tri shared mem > device limits | ✅ Limit captured on `VulkanDevice`, guard deferred to PR2 (documented) |
| #2 conv1d dropped conv_state | ✅ Gated off by default, full fix + parity test deferred to PR2 |
| #3 probe() instance leak | ✅ `instance.destroy_instance(None)` before success return |
| #3 runtime compile writes source tree | ✅ `compile_shader_runtime` uses `std::env::temp_dir()` |
| #3 build.rs invisible `println!` | ✅ Silent on success, `cargo:warning` only for real problems |
| #3 silu overflow in causal_conv1d | ✅ Sign-branched stable silu applied |
| #3 remaining inline pipelines | 📋 Deferred (3 sites, structural reason for `dispatch_kernel`, two foldable in PR2) |
| #3 document FORWARD_SUB kill-switch | ✅ Documented in `supports_gdn_forward_substitution` and PR body |

I'm comfortable with this PR landing once two small things are tidied. Both are tiny diffs and neither is a hard blocker — the call is yours on whether to do them here or open them as PR2's first commit.

---

### 1. Env-var-read-per-call regressed; the cached-flags comment is now wrong

The struct comment in `VulkanBackend` still says (`vulkan.rs:26-27`):

> *Cached at construction: reading env vars per decode step × 24 GDN layers shows up in decode NVTX captures. Env vars don't change at runtime.*

That comment was written *because* of a real perf finding. But the round-3 commit removed `fused_conv1d_enabled` from the struct's cached fields and replaced its uses with inline `std::env::var("KILN_ENABLE_VULKAN_FUSED_CONV1D").is_ok()` calls. That call is now made:

- In `supports_causal_conv1d_update` (vulkan.rs:194)
- In `supports_causal_conv1d_prefill` (vulkan.rs:201)
- In `causal_conv1d_update` (the impl body, vulkan.rs:418)
- In `causal_conv1d_prefill` (the impl body, vulkan.rs:443)

Per decode token × 24 GDN layers, that's ~96 `std::env::var` calls just for conv1d gating, plus another 48 for `supports_gdn_forward_substitution` which also reads two env vars per call (`KILN_DISABLE_VULKAN_GDN_FORWARD_SUB` + `KILN_ENABLE_VULKAN_GDN_FORWARD_SUB`, vulkan.rs:162-163). `std::env::var` on Linux locks the process-global env table — not free at this rate.

Fix: move both back into the cached struct fields, alongside the existing `gdn_*_enabled` flags. Pattern matching the existing convention:

```rust
pub struct VulkanBackend {
    // ...
    fused_conv1d_enabled: bool,
    gdn_forward_sub_enabled: bool,
    // ...
}

impl VulkanBackend {
    pub fn new(device: Device) -> Self {
        // ...
        let fused_conv1d_enabled = std::env::var("KILN_ENABLE_VULKAN_FUSED_CONV1D").is_ok();
        let gdn_forward_sub_enabled = std::env::var("KILN_DISABLE_VULKAN_GDN_FORWARD_SUB").is_err()
            && std::env::var("KILN_ENABLE_VULKAN_GDN_FORWARD_SUB").is_ok();
        // ...
    }
}
```

Then `supports_*` and the impl bodies read the cached bool. Default-off semantics are preserved.

### 2. Dead `let _ = std::env::var(...)` line

`vulkan.rs:46` has:

```rust
let _ = std::env::var("KILN_DISABLE_FUSED_CONV1D"); // backward compat
```

`let _ =` discards the result — this isn't doing anything. If you want to honor the legacy env var (warn users that it's been replaced), it should be:

```rust
if std::env::var("KILN_DISABLE_FUSED_CONV1D").is_ok() {
    tracing::warn!(
        "KILN_DISABLE_FUSED_CONV1D is no longer used; \
         fused conv1d is now opt-in via KILN_ENABLE_VULKAN_FUSED_CONV1D"
    );
}
```

If you don't actually care about backward compat (no users of this env var exist in the repo or downstream), just delete the line.

### 3. `probe()` still leaks the instance on the error path

The success path now correctly destroys the instance:

```rust
let available = !devices.is_empty();
unsafe { instance.destroy_instance(None); }
available
```

But the `enumerate_physical_devices` failure path returns early without destroying:

```rust
let devices = match unsafe { instance.enumerate_physical_devices() } {
    Ok(d) => d,
    Err(_) => return false,  // ← instance still alive
};
```

Easiest fix is the same as the success path — wrap in a closure or use a struct-with-Drop:

```rust
let result = unsafe { instance.enumerate_physical_devices() };
let available = result.map(|d| !d.is_empty()).unwrap_or(false);
unsafe { instance.destroy_instance(None); }
available
```

Vulkan instances are small and probe is one-shot per process, so this is genuinely tiny — but easy to fix while you're in there.

---

### Smaller things, take or leave

- **`max_compute_shared_memory_size` is `vk::DeviceSize` (u64)**, but the underlying `VkPhysicalDeviceLimits` field is `uint32_t`. Either is fine; for symmetry with the source type and to make the API obvious, `u32` is more truthful. Not a real issue.
- **`supports_gdn_forward_substitution` doc is good**, but consider also putting the `KILN_ENABLE_VULKAN_FUSED_CONV1D` semantics in the same docstring style on `supports_causal_conv1d_update` so anyone running `grep -r "KILN_ENABLE" crates/kiln-model` finds the explanation in one place.
- **The `VulkanDevice` `Debug` impl** now omits `max_compute_shared_memory_size`. Nice to add to the debug print so startup logs include "shared mem 32768 bytes" alongside vendor/device — diagnostic gold for the PR2 dispatch guard.

---

### Recommended state to merge

If you fix items 1 and 2 in this PR (item 3 is genuinely small enough to leave as-is) it's good to land. Both are < 30-line diffs. If you'd rather flip immediately to PR2 and pick these up in its first commit, that's also reasonable — they're not user-facing bugs at the merged-but-unused state of this backend.

The cumulative PR (`-1,901` LOC, draft, mergeable, four iterations) is a solid scaffold to build on. Two unit tests now in the new crate (the device probe smoke test and the shader dispatch test). Build/test commands in the body still match what we'd run in CI.

---

### PR2 framing

Carrying the queue forward, here's what the PR2 task list now looks like, in order of dependency:

1. **`KilnVulkanStorage` design + implementation.** Discuss before code — three viable shapes from the round-3 review (newtype above candle / side-table over Tensor / wait for upstream candle Vulkan). This is the architectural keystone and the right hour-long whiteboard before any code lands.
2. **Wire `ShaderPipeline::get_or_create` into `VulkanBackend`.** Pre-create all 9 pipelines at backend construction; dispatch path looks them up by stable key. Drops the current per-call create+destroy churn.
3. **Add the dispatch-time guard for `solve_tri`** using `VulkanDevice::max_compute_shared_memory_size()`. Decline the kernel and return `Ok(None)` if shared-mem footprint would exceed the device limit. Document the threshold computation alongside the constant `C ≤ 128, dv ≤ 256` doc.
4. **Fix `causal_conv1d` properly**: rebind `conv_state` as binding 2 in the output kernel; rewrite `state_advance.comp` to handle `T < K-1` (combine input tail with old-state snapshot) and `T ≥ K-1` (copy from input tail directly). Land alongside its parity test.
5. **CPU-parity harness for `gdn_gates`** as the first vertical-slice kernel. Random fuzzed inputs, max-abs-diff < 1e-3 vs candle reference. The harness will be reused for the conv1d fix above and every kernel after.
6. **Decode bench on real AMD hardware** with `gdn_gates` only on the Vulkan path, all other GDN kernels off (or via candle). Compare to CPU-only baseline. Bench numbers in the PR body.

If `gdn_gates`-only on the Vulkan path doesn't show neutral-or-better tok/s vs CPU-only baseline, *that's the data point* that says "the storage architecture as designed isn't enough; need to migrate matmuls to live on the GPU before more kernels are worth shipping" — and we redesign before PR3+ kicks off. Getting that signal early is the most valuable thing PR2 produces.

---

Four rounds in, the workflow is humming. Every item from every prior review is tracked, every fix is verifiable from the diff, every deferral is explicit and justified. That discipline is the best forward-looking signal we have for whether this whole effort hits — keep it going into PR2 and the parity-test bar I asked for ("every kernel ships with its parity test in the same PR") and we'll know in one or two more iterations whether the architecture is going to pay out. Nice work.

ericflo:
## Round-5 review — PR1 commit `573ffbd`

This iteration is a clean conclusion to PR1. Every round-4 item resolved without regressions:

| Round-4 item | Status |
|---|---|
| #1 env-var-per-call regression | ✅ `fused_conv1d_enabled` and `gdn_forward_sub_enabled` restored as cached struct fields. All 5 use sites (2 `supports_*`, 1 `supports_gdn_forward_substitution`, 2 impl bodies) read the cached bool. The "Cached at construction" comment is now consistent with reality. ~144 `std::env::var` calls per decode step eliminated. |
| #2 dead `let _ = std::env::var(...)` line | ✅ Removed and subsumed into the `gdn_forward_sub_enabled` initialization, with a clearer block comment explaining why both opt-ins exist. |
| #3 `probe()` leak on error path | ✅ Unified: `enumerate_physical_devices().map(\|d\| !d.is_empty()).unwrap_or(false)` followed by unconditional `destroy_instance`. Both success and error paths now clean up. Doc updated to "explicitly destroyed on every path." |
| Bonus: Debug impl missing field | ✅ `max_compute_shared_memory_size` is in the `VulkanDevice` Debug output now, so startup logs and crash dumps surface the per-device shared-memory budget. |

The diff is tight (`+31 / −34`), purely cleanup, and reads cleanly top-to-bottom. The struct comment about cached env vars is back to being load-bearing rather than aspirational, which is the thing I most wanted to see.

---

### One small observation, not blocking

The cached `gdn_forward_sub_enabled` checks `gdn_enabled && KILN_ENABLE_VULKAN_GDN_FORWARD_SUB`, but `fused_conv1d_enabled` checks only `KILN_ENABLE_VULKAN_FUSED_CONV1D` (no `gdn_enabled` gate). Previously both ungated paths bypassed the master `KILN_DISABLE_GDN_KERNEL` switch — now solve_tri respects it but conv1d doesn't.

Argument for the asymmetry: depthwise causal-conv1d is technically part of the GDN block in Qwen3.5 but is wired through its own kernel-trait method, so treating it as separate from the GDN master switch is defensible.

Argument against: a user who sets `KILN_DISABLE_GDN_KERNEL=1` (e.g., to debug a parity issue against the candle path) will still get the experimental Vulkan conv1d if they had `KILN_ENABLE_VULKAN_FUSED_CONV1D` set. That's a small footgun.

Easiest path: gate `fused_conv1d_enabled` on `gdn_enabled` too (the conv lives inside the GDN block in practice), so the master switch turns off everything GDN-block. Two-line change. Take it or leave it — neither side bites a real user today since the kernel is off by default.

---

### Status: ready to merge

I think this PR is in shape to land:

- All correctness blockers from rounds 1–4 are either fixed in-tree or gated behind explicit opt-ins with documented PR2 follow-ups.
- The architectural deferrals (`KilnVulkanStorage`, pipeline cache wiring) are documented in the module docs and PR body, not papered over.
- The hygiene work is done: 1,950 net LOC removed, 14 dead shaders gone, 12 scratch examples gone, `*.spv` ignored, `PLAN-VULKAN.md` moved under `docs/audits/`.
- Two unit tests in the new crate (`device::tests` + `examples/dispatch_test.rs`).
- Mergeable, draft, no merge conflicts.

When you're ready, mark the PR ready-for-review and we can land it. No further round needed unless something surfaces in CI.

---

### PR2 priorities (carrying forward, in dependency order)

1. **`KilnVulkanStorage` design discussion** before any code lands. Three viable shapes still on the table; pick one.
2. **Wire `ShaderPipeline::get_or_create` into `VulkanBackend`** — pre-create all 9 pipelines at backend construction, dispatch path looks them up by stable key.
3. **Dispatch-time shared-memory guard for `solve_tri`** using `VulkanDevice::max_compute_shared_memory_size()`. Decline to `Ok(None)` if the kernel won't fit.
4. **Fix `causal_conv1d` properly**: re-bind `conv_state` as binding 2 in the output kernel; rewrite `state_advance.comp` to handle `T < K-1` and `T ≥ K-1` correctly. Land alongside its parity test.
5. **CPU-parity harness for `gdn_gates`** — first vertical-slice kernel. Random fuzzed inputs, max-abs-diff < 1e-3 vs candle. The harness extends naturally to the conv1d fix above and every kernel after.
6. **Decode bench on real AMD hardware** with `gdn_gates` only on the Vulkan path. Compare to CPU-only baseline. Bench numbers in the PR body.

Item 6 is the one to optimize for: it's the data point that decides whether the architecture needs a wider redesign before PR3+ scales out the rest of the kernels. Get it early and unambiguous.

Non-negotiable carryover from prior rounds: every kernel in PR2+ ships with its parity test in the same PR. The conv1d regression in round-3 is the canonical illustration of why — it shipped functionally broken because there was no parity test gating it. That bar converts "I think this works" into "the test asserts it works" and is the highest-leverage thing we can hold the line on.

---

Five rounds in: every prior finding tracked, every fix verifiable from the diff, every deferral explicit and justified, no regressions on the cleanup commits. The discipline is the model — keep it going into PR2 and we'll converge on the architecture-pays-out signal in one or two more iterations. Nice work.

PR #638, opened 2026-04-29
phase10: Liger-style RMSNorm with custom CUDA backward (kernel-fusion pass §1)

ericflo:
## What

Phase 10 §1: Liger-style fused RMSNorm with a custom CUDA backward kernel.

- New CUDA kernel `fused_rmsnorm_bwd.cu` (204 lines): single-pass per-row reduction for grad_x and grad_w using shared memory + atomicAdd.
- New `CustomOp2` wrapper in `kiln-rmsnorm-kernel` exposing forward+backward to candle's autograd graph.
- Dispatch in `forward.rs`: training path goes through the fused custom op; `KILN_DISABLE_RMSNORM_BACKWARD=1` falls back to the candle op chain (non-fused).
- Trainer parity test: SFT step on CPU is bit-exact between fallback paths (sanity check that the dispatch logic is wired correctly).
- Kernel parity test on CUDA: decode-row exact, multi-row within 2e-2 (atomicAdd ordering + F32→bf16 cast yields ~1 bf16 ULP variance over 512 row reductions).
- Microbench: kernel-only forward+backward sweep at Qwen3.5-4B hidden=2560, BF16, A6000.

## Why

Phase 10 §1 of the kernel-fusion pass. Liger fuses normalization layers with their backward to reduce memory traffic at long contexts. This kernel removes the candle-op-chain backward (which produces multiple intermediate tensors) and replaces it with a single fused reduction. Direct enabler for §2 (RoPE) and §3 (SwiGLU) which follow the same pattern.

## Bench

Kernel-only `fwd + sum + backward` microbench (no model load, no trainer overhead), BF16 on A6000, hidden=2560, batch=1, median of 25 iters after 5 warmup. Compares the fused custom-op path against the candle op-chain path (the `KILN_DISABLE_RMSNORM_BACKWARD=1` baseline).

| shape              |    T | candle ms | fused ms | speedup |
|:-------------------|-----:|----------:|---------:|--------:|
| decode             | 1    |     0.444 |    0.123 | **3.61×** |
| training T=512     | 512  |     2.695 |    1.417 | **1.90×** |
| training T=2048    | 2048 |    11.674 |    5.523 | **2.11×** |
| training T=4096    | 4096 |    23.191 |   11.588 | **2.00×** |
| training T=8192    | 8192 |    48.503 |   23.792 | **2.04×** |

A clean 2× across all training shapes — both saved-tensor reduction (fwd skips the F32 intermediates the candle chain materializes) and launch-count reduction (single fused fwd kernel + single fused bwd kernel vs ~11 candle launches each direction). Decode is even better (3.6×) because the per-call overhead dominates at T=1.

End-to-end SFT bench (`phase10_rmsnorm_bench`, T=2048/4096/8192) was attempted but blocked: the W4A16+SFT path errors with `expected rank-2 q_proj_t for layer 3` (Marlin-packed 1D tensor, pre-existing, out of scope for this PR), and without W4A16 the bench OOMs at the 48 GiB ceiling on A6000 even with FLCE on at T=2048. Microbench captures the per-kernel speedup directly without those confounders. End-to-end peak-VRAM verification can ride into a follow-up once the W4A16+SFT trainer compatibility is fixed.

## Tests

- `cargo test --release -p kiln-rmsnorm-kernel` — 13 unit tests pass (CPU + CUDA parity for fwd/bwd, decode-row exact, multi-row within tolerance)
- Trainer SFT parity test passes on CPU between fused path and `KILN_DISABLE_RMSNORM_BACKWARD=1` fallback (bit-exact loss)

## Phase 10 progress

- [x] §1 RMSNorm with custom backward (this PR)
- [ ] §2 Fused RoPE
- [ ] §3 Fused SwiGLU
- [ ] §4 Fused Layer Norm
- [ ] §5 FleCE chunked CE

### DIFF
--- a/crates/kiln-model/src/forward.rs
+++ b/crates/kiln-model/src/forward.rs
@@ -1803,18 +1803,33 @@ fn embedding_lookup_from_transposed(token_ids: &[u32], embed_tokens_t: &Tensor) 
 ///
 /// Returns: same shape as `x`.
 ///
-/// On CUDA+bf16 inputs within the kernel envelope (hidden <= 8192), dispatches
-/// to `kiln_rmsnorm_kernel::fused_rmsnorm`, which collapses the ~11 candle op
-/// launches (to_dtype, sqr, mean_keepdim, +eps, sqrt, recip, broadcast_mul,
-/// to_dtype, ones_like + w, broadcast_mul, to_dtype) into a single fused kernel.
-/// Falls back to the candle-op path on CPU, non-bf16, or out-of-envelope inputs.
+/// Dispatch (CUDA path; bf16 inputs within the kernel envelope, hidden <= 8192):
+///   - Default: `kiln_rmsnorm_kernel::fused_rmsnorm_with_autograd` — fused
+///     forward kernel + manual CUDA backward kernel routed through
+///     `CustomOp2`. The autograd graph saves only `x` and `weight` (not the
+///     F32 intermediates that the candle-op chain materializes), shrinking
+///     per-layer saved-tensor peak during Phase 10 long-context training.
+///   - `KILN_DISABLE_RMSNORM_BACKWARD=1`: `kiln_rmsnorm_kernel::fused_rmsnorm`
+///     forward only, autograd-traced backward (debug fallback for isolating
+///     fwd-vs-bwd divergence).
+///   - `KILN_DISABLE_RMSNORM_KERNEL=1`: full `rms_norm_fallback` candle-op
+///     chain (kill switch).
+///
+/// On CPU, non-bf16, out-of-envelope, or non-CUDA backends: falls back to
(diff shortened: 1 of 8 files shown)


ericflo:
CI failures (`test_layer_pair_tiled_matches_monolithic_cpu_hybrid` Linux + `test_checkpointed_forward_backward_tiled_matches_monolithic_cpu` macOS) are pre-existing on main from PR #637's merge — same NaN panics, same line numbers, same tests, on commit c983ca7 (run [25108241466](https://github.com/ericflo/kiln/actions/runs/25108241466)). Both tests are in tiling code added by recent PRs and are unrelated to the RMSNorm custom-backward changes here. Merging with admin override; the broken tests should be fixed in a separate PR.

PR #976, opened 2026-05-06
docs(readme): surface model weights download step before Docker block

ericflo:
## What

Extract the `huggingface-cli download Qwen/Qwen3.5-4B --local-dir ./Qwen3.5-4B` step into a single shared **Get the model weights** subsection that appears before Path 2, so a cold reader trying any of Paths 2–4 (server binary, container, source build) sees the weights step before the run command. Removes the duplicate copies that previously lived inside Path 2 and Path 4. Path 3 (Docker) now references the resulting `./Qwen3.5-4B` directory directly via `-v "$(pwd)/Qwen3.5-4B:/models/Qwen3.5-4B:ro"` instead of the placeholder `/path/to/Qwen3.5-4B`.

## Why

Cold-reader friction in Path 3 (Container): the `docker run` block told users to "Replace `/path/to/Qwen3.5-4B` with your local `Qwen3.5-4B` model directory" but never explained how to obtain those weights — the download command only appeared inside Path 2's code block and again in Path 4. A user who skipped to Docker first had to scan two other paths to figure out where the weights came from. Path 1 (Desktop App) is unaffected — the app still handles weights automatically.

## Verification

- README diff reads cleanly cold: weights-download command now visible before the docker run example.
- `huggingface-cli download Qwen/Qwen3.5-4B --local-dir ./Qwen3.5-4B` appears exactly once in the README (`grep -c` returns `1`).
- 13 insertions, 13 deletions — well under the 25-line cap.
- No Rust changes; no other files touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ huggingface-cli download Qwen/Qwen3.5-4B --local-dir ./Qwen3.5-4B
 
 This downloads `Qwen/Qwen3.5-4B` into `./Qwen3.5-4B`, which the commands below reference directly.
 
-**Path 2 — Server binary (terminal-first, no source build):** Download the latest `kiln-v*` server artifact when you want the `kiln` server in your terminal with no source build, Desktop App, or Docker. Linux x86_64 + NVIDIA CUDA 12.4 is the compact path:
+**Path 2 — Server binary (terminal-first, no source build):** Download the latest `kiln-v*` server artifact when you want the `kiln` server in your terminal with no source build, Desktop App, or Docker. Linux x86_64 + NVIDIA CUDA 12.4 is the compact path (run the `Qwen/Qwen3.5-4B` weights step above first):
 
 ```bash
 KILN_VERSION=$(curl -fsSL https://api.github.com/repos/ericflo/kiln/releases/latest | sed -n 's/.*"tag_name": "kiln-v\([^"]*\)".*/\1/p')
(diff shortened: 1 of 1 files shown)


ericflo:
CI failure on this PR is **pre-existing main-branch drift, unrelated to the README change**.

**Failure:** `crates/kiln-server/src/cli.rs: SERVE_OVERVIEW missing QUICKSTART.md` (from `scripts/check_docs_site_smoke.mjs`)

**Root cause:** PR #973 (`cli: replace local-only QUICKSTART.md refs with canonical docs URLs`) replaced the literal `QUICKSTART.md` references in `cli.rs` with `https://ericflo.github.io/kiln/quickstart.html`, but `scripts/check_docs_site_smoke.mjs` (lines 915 and 930) still requires the literal `QUICKSTART.md` string in `SERVE_OVERVIEW` and `HEALTH_OVERVIEW` constants.

**Why it wasn't caught on #973:** The `Release Version Drift` workflow path filter in `.github/workflows/release-version-drift.yml` does not include `crates/kiln-server/src/cli.rs`, so the smoke test never ran against #973's diff. My README change is the first edit to a watched path since #973 landed.

**Scope:** This task is constrained to `README.md` only. The smoke test ↔ cli.rs drift should be a separate one-line PR (either drop the `QUICKSTART.md` entries from `SERVE_OVERVIEW`/`HEALTH_OVERVIEW` in the smoke test, or add the path-filter for `cli.rs` and reconcile). Surfacing here so the next planning cycle can queue it.

The README change itself is intact and verified: `huggingface-cli download Qwen/Qwen3.5-4B --local-dir ./Qwen3.5-4B` appears exactly once, all four Quick Start paths now reference the shared "Get the model weights" step, and Path 3 Docker `-v` flag uses `$(pwd)/Qwen3.5-4B`.

PR #1070, opened 2026-05-22
cuda_sft_file: default --trainer to generic; hoist native trainer allocs (#1063)

### LINKED ISSUE #1063: cuda_native_sft_train is 50× slower than --trainer generic — backport vk_native_sft_train structure
## Problem

`cuda_native_sft_train` runs at ~180 s/step on Qwen3.5-4B on an A6000 while `--trainer generic` (which calls `sft_train` in `trainer.rs`) runs at ~3.5 s/step. **~50× speed gap** for the same model, same dataset, same LoRA rank.

Reproduced 2026-05-22 on pi-faithful-completion iter6 (rank 4, lr 1e-5, 211 examples). The training itself does converge correctly when given enough time — but the gap is so large that nobody should pick the native trainer in practice today.

## Diagnosis

In `crates/kiln-train/src/cuda_train.rs:1325` (`cuda_native_sft_train`):

1. **Per-step `CudaTrainArena::new()`** (line 1469) — allocates a fresh CUDA arena every step rather than hoisting outside the loop.
2. **Per-step `cuda_linear_attention_state_zeros_for_model(&model, 1)`** (line 1471) — re-allocates the GDN state on every step instead of allocating once and zeroing in place.
3. **The step kernel `cuda_lora_model_adamw_step_with_gdn_state_with_arena`** does many small CUDA launches with CPU sync between them; `nvidia-smi -l 1` shows GPU util near 0% while the CPU dispatches.

Compare `vk_native_sft_train` in `vk_train.rs:4706`, which has:

- gradient checkpointing (`vk_checkpointed_train_step`) with configurable segments
- exact layerwise recompute for hybrid GDN (`vk_recompute_train_step_with_state_masked`)
- length-sorted iteration (`KILN_VK_LENGTH_SORT_SFT`)
- pre-computed GDN state shape outside the loop
- three distinct fast paths (ckpt / recompute / plain) chosen by model shape

## Proposed fix

Backport the structure from `vk_native_sft_train` to `cuda_native_sft_train`:

1. Hoist `CudaTrainArena::new()` and `cuda_linear_attention_state_zeros_for_model` out of the for loop; reset in place.
2. Add gradient checkpointing path matching `vk_checkpointed_train_step`.
3. Route the inner step through `BackendRuntime` (the same dispatcher `sft_train` uses), inheriting fused-kernel paths.

Easier interim: change `let mut trainer = TrainerKind::Native;` (cuda_sft_file.rs:86) to default to `Generic`, and rename `Native` → `NativeExperimental` to discourage use.

## Why this matters

Every capability that uses cuda_sft_file pays this tax by default today. The pi-faithful-completion iter23 ship (round-3) took ~75 min on generic; on native it would take ~63 hours.

ericflo:
Closes #1063 (the user-visible 50× slowdown). Tracks the deeper
BackendRuntime backport as follow-up.

### User-visible fix (the 50× win for all callers today)
- `cuda_sft_file` now **defaults to `--trainer generic`**. The
  legacy CUDA-native trainer (`cuda_native_sft_train`) was the silent
  default and is currently ~180 s/step on Qwen3.5-4B (rank 4, lr
  1e-5) vs ~3.5 s/step for the generic trainer that routes through
  `sft_train` -> `BackendRuntime`.
- The experimental CUDA-native path is now exposed as
  `--trainer native-experimental`. `--trainer native` is retained as
  a legacy alias that maps to the same path so existing scripts and
  capabilities do not break overnight, but it now prints a clear
  startup warning citing this issue.
- `cuda_sft_file --help` documents the trainer choices, marks
  `generic` as the recommended/default path, and labels
  `native-experimental` as ~50× slower today.

### Structural backport from `vk_native_sft_train` (per the issue body)
- `cuda_native_sft_train` hoists per-step `CudaTrainArena::new()` and
  `cuda_linear_attention_state_zeros_for_model()` out of the inner
  loop. The arena is reused via `CudaTrainArena::clear()` and the
  GDN state is reset via the new
  `CudaLinearAttentionState::zero_in_place()` method.
- `cuda_lora_train_token_sequences_with_gdn_state` and
  `cuda_full_attention_lora_train_token_sequences` get the same
  hoist for consistency.
- New `KILN_CUDA_LENGTH_SORT_SFT` env tristate enables length-sorted
  iteration, mirroring `KILN_VK_LENGTH_SORT_SFT`. Default heuristic:
  auto-on when any example is at least 8K tokens AND there is more
  than one example to reorder. Sort changes step order only; the
  trainer never truncates.
- Per-step `info!` now logs `seq_len`, `original_index`, and
  `step_ms` so the slow path is diagnosable from a single tail
  (matches `vk_native_sft_train`'s log shape).
- Startup `warn!` on entry into `cuda_native_sft_train` re-states the
  perf gap and links this issue, so anyone watching logs sees the
  slowdown explained rather than as silent degradation.

### Test
- New unit test
  `cuda_linear_attention_state_zero_in_place_resets_layer_state`
  verifies zero/shape/dtype/metadata preservation and idempotence
  across two consecutive resets.

## What this does *not* change

These hoists close the malloc/zeros pressure flagged in the issue
diagnosis (items 1–2) but do **not** close the 50× gap on their own —
the remaining gap is the small-launch + CPU-sync structure of the
inner step kernel (item 3, "BackendRuntime routing"). Closing item 3
is a substantially larger refactor that should land as its own PR
once the path is scoped. The default flip is what eliminates the 50×
gap for all `cuda_sft_file` users today, exactly as the issue's
"easier interim" recommends.

## Build / validation

I was unable to run `cargo build --release --features cuda` on a real
A6000 for this PR: RunPod A6000 capacity was unavailable across 6
acquire attempts (HTTP 500 "no longer any instances available with
the requested specifications"). Per the `runpod-capacity-fallback`
pattern I'm shipping the code and flagging the validation gap so a
follow-up smoke test can land once capacity is back.

The changes are deliberately conservative for that reason:

- `cuda_sft_file.rs` changes are a default-value flip plus a renamed
  variant; no kernel changes.
- The arena hoist is semantically equivalent: `let mut arena =
  CudaTrainArena::new(...)` inside the loop drops the arena (and its
  tracked tensors) at end of iteration; `arena.clear()` at the top of
  the next iteration explicitly drops the same `Vec<CudaTrainTensor>`
  before the next step begins. The CUDA caching allocator behaves
  identically.
- `zero_in_place` re-creates the same-shape/same-dtype zero tensors
  on the same device — bit-for-bit equivalent to the previous
  per-step `cuda_linear_attention_state_zeros_for_model` call.
- Length-sort defaults to OFF (`max_seq_len < 8192 || tokenized.len()
  <= 1`), so the default training path for small/short SFT datasets
  is identical to before.

Suggested smoke test once A6000 is available:

```bash
# 1) Build verifies the new TrainerKind variant + zero_in_place compile.
KILN_CUDA_ARCHS=86 cargo build --release --features cuda --example cuda_sft_file

# 2) Library tests cover the new zero_in_place behavior.
KILN_CUDA_ARCHS=86 cargo test --release --features cuda -p kiln-model \
  cuda_linear_attention_state_zero_in_place

# 3) Smoke that the new default works end-to-end on a tiny SFT slice:
./target/release/examples/cuda_sft_file \
  --data examples.jsonl --model-path /workspace/qwen3.5-4b \
  --output-dir /workspace/tmp/sft --adapter-name smoke \
  --epochs 1 --max-examples 4
# expect: "trainer=generic" in the startup banner, NO 50×-slowness
# warning, total wall < 1 min on 4 short examples.

# 4) Confirm the legacy alias still routes (and warns) into the
#    experimental path:
./target/release/examples/cuda_sft_file --trainer native \
  --data examples.jsonl --model-path /workspace/qwen3.5-4b \
  --output-dir /workspace/tmp/sft-legacy --adapter-name smoke-legacy \
  --epochs 1 --max-examples 2
# expect: "trainer=native-experimental" in the startup banner, the
# new WARNING line printed to stderr, the
# trainer_warning=cuda_native_experimental_50x_slower line on stdout.
```

## Files

- `crates/kiln-train/examples/cuda_sft_file.rs` — default flip,
  rename, warning, help text.
- `crates/kiln-train/src/cuda_train.rs` — hoist arena + gdn_state in
  three functions, add length-sort tristate, add per-step tracing,
  add startup warn on the experimental path.
- `crates/kiln-model/src/cuda_train.rs` — new
  `CudaLinearAttentionState::zero_in_place` + unit test.
- `.agents/skills/capability-creator/resources/sft-mode.md` —
  capability-creator guidance updated for the default flip.
- `CHANGELOG.md` — Unreleased entry.

Auto-merge once Eric (or anyone reviewing) is satisfied with the
shape. Per kiln convention this PR has no `CE:` prefix and there is
no deploy-request — kiln does not have CI today, so `gh pr checks`
will report empty.

### DIFF
--- a/.agents/skills/capability-creator/resources/sft-mode.md
+++ b/.agents/skills/capability-creator/resources/sft-mode.md
@@ -202,11 +202,16 @@ applied to a different cap would pick different distributions.
 
 ## Practical gotchas
 
-- **Use `--trainer generic`.** The native trainer
-  (`cuda_native_sft_train`) is currently ~50× slower than generic on
-  Qwen3.5-4B; the generic path (`sft_train` in `trainer.rs`) routes
-  through `BackendRuntime` and gets the production-tuned kernels. See
-  kiln issue #1063 for the backport tracking the native trainer.
+- **Use `--trainer generic` (now the default).** As of the kiln issue
+  #1063 fix, `cuda_sft_file` defaults to `--trainer generic` so the
+  CLI no longer silently picks the slow path. The legacy CUDA-native
+  trainer (`cuda_native_sft_train`) is exposed as
+  `--trainer native-experimental` and is currently ~50× slower than
+  generic on Qwen3.5-4B; the generic path (`sft_train` in
+  `trainer.rs`) routes through `BackendRuntime` and gets the
+  production-tuned kernels. Existing capabilities that still pass
+  `--trainer generic` continue to work — leaving the flag in is
+  defensible and a clearer signal of intent.
 - **Flatten the adapter directory after training.** `cuda_sft_file`
   writes to `<output-dir>/<adapter-name>/...` but kiln serve expects
   `<output-dir>/...` directly. After training, `mv` the inner files up
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,67 @@
 # Kiln Server Changelog
 
+## Unreleased — cuda_sft_file defaults to --trainer generic (issue #1063)
+
+`cuda_sft_file` no longer silently picks the ~50× slower CUDA-native
+SFT path. The legacy CUDA-native trainer (`cuda_native_sft_train`) is
+exposed as `--trainer native-experimental` and runs ~180 s/step on
+Qwen3.5-4B (rank 4, lr 1e-5) vs ~3.5 s/step for the generic trainer
+(`sft_train` -> `BackendRuntime`); the default now matches the
+recommended workflow used by `caps/pi-faithful-completion` and other
+production SFT capabilities.
+
+### Changed
+- `cuda_sft_file` `--trainer` now defaults to `generic` rather than the
+  experimental CUDA-native path. New CLI value
+  `--trainer native-experimental` opts into `cuda_native_sft_train`;
+  the legacy `--trainer native` is accepted as an alias for backwards
+  compatibility but now routes to the same experimental path and
+  prints a clear startup warning citing kiln issue #1063.
+- `--help` text on `cuda_sft_file` now documents the trainer choices,
+  marks `generic` as the recommended/default path, and labels
+  `native-experimental` as ~50× slower today.
+
+### Performance
+- `cuda_native_sft_train` (the experimental path) hoists the
+  per-step `CudaTrainArena::new()` and
+  `cuda_linear_attention_state_zeros_for_model()` allocations out of
+  the inner training loop. The arena is now reused via
+  `CudaTrainArena::clear()` per step, and the GDN state is reset via
+  the new `CudaLinearAttentionState::zero_in_place()` method.
+  `cuda_lora_train_token_sequences_with_gdn_state` and
+  `cuda_full_attention_lora_train_token_sequences` receive the same
+  hoist. These changes close the malloc/zeros pressure in
+  cuda_native_sft_train but do not close the 50× gap on their own —
+  the remaining gap is the small-launch + CPU-sync structure of the
+  inner step kernel, tracked as future work on the issue.
+- `cuda_native_sft_train` now supports length-sorted iteration via
+  the new `KILN_CUDA_LENGTH_SORT_SFT` env tristate, mirroring the
+  Vulkan side's `KILN_VK_LENGTH_SORT_SFT`. Default heuristic: enable
+  when any example is at least 8K tokens AND there is more than one
+  example to reorder. Sort changes step order only; the trainer never
+  truncates.
+- `cuda_native_sft_train` now logs `seq_len`, `original_index`, and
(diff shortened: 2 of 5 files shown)


ericflo:
Closing in favor of #1071, which implements the structural backport from the issue body's 'Proposed fix' section 3 (layerwise reverse-recompute) rather than this PR's interim default flip. The hoisted alloc + length-sort changes from this PR are not carried over because the recompute step replaces the legacy step's setup entirely and pays its own boundary-cache cost; the default flip itself isn't needed either because #1071 keeps the recompute path opt-in (`KILN_CUDA_RECOMPUTE_SFT=1`) after empirically confirming that recompute alone does not close the 50× gap. The real gap is per-op CPU overhead in `cuda_train`'s autograd, which #1071's bench section discusses.

PR #1361, opened 2026-05-24
phase 7: migrate all 20 gdn-kernel kt-API entries to owner-agnostic ptrs

ericflo:
Final Phase 7 v2 kt-API migration batch. All 5 kernel crates now at 100% Borrowed-compatibility.


ericflo:
Reverting — automated script left functions in a broken state (cuda_output_device_ptr matched as a _ptr name, mapping by order failed for multi-output functions). Will redo with smaller hand-crafted batches.

PR #1413, opened 2026-05-29
kiln-train: tape-authoritative SFT backward + substrate gate (#1082 CP-4 Step 2)

ericflo:
## What

Wires the **trainer flip** for #1082 CP-4 (the endgame). With `KILN_TAPE_AUTHORITATIVE=1`, `standard_forward_backward` now routes SFT backward through the kt `Tape` (seeded from the cross-entropy loss) instead of candle's `loss.backward()`.

This is **Step 2** of the trainer flip — it lands on top of Step A (kt-id chaining primitives) and Step B (`with_tape_authoritative_scope`, `31208400` / `fb0c2f30`).

**Default behavior is unchanged**: the gate is off by default and the existing tape-bridge / candle paths still run on every other config.

## Changes

1. **`kiln-autograd`** — new `tape_authoritative_enabled()` reading `KILN_TAPE_AUTHORITATIVE` (OnceLock-cached, default off). Parallel in shape to `tape_forward_enabled()` / `KILN_USE_TAPE_FORWARD`.

2. **`kiln-train::cross_entropy_loss`** — attempts `try_tape_cross_entropy_cuda(active_logits_f32, labels_tensor)` before the existing composite. When the adapter records (CUDA + tape scope active + canonical shapes), the returned loss is a kt-mirror with `CrossEntropyKtBackward` on the tape; `with_tape_authoritative_scope` then uses it as the `dL/dL = 1` seed. Falls through transparently on CPU / no scope / out-of-envelope.

3. **`kiln-train::standard_forward_backward`** — gate-dispatches into the new `standard_forward_backward_tape_authoritative` when `KILN_TAPE_AUTHORITATIVE=1`. The new function:
   - Requires `KILN_USE_TAPE_FORWARD=1` (per-op adapters must record onto the tape).
   - Requires `KILN_USE_FLCE=0` (FLCE's adapter doesn't register a candle⟷kt output mapping yet — separate follow-up).
   - Wraps `model_forward` + `cross_entropy_loss` in `with_tape_authoritative_scope`.
   - Builds the candle `GradStore` via **option (a)** from `kiln-cp4-trainer-flip-gradstore-wrinkle`: `loss.backward()` on the detached candle loss returns a near-empty store (just `loss → ones_like(loss)` — the upstream graph is empty because the loss is a fresh kt-copy), then `insert_id` each tape-derived per-Var grad keyed by `var.as_tensor().id()`.
   - Emits a `tracing::warn!` (does not abort) when zero LoRA Vars matched the tape's registered candle inputs, so a user opting into the flag without the LoRA-path adapter follow-up sees the foot-gun clearly in logs.

4. **Parity test (substrate gate)** — `cuda_tape_authoritative_grads_match_standard_tiny` exercises the tiny_config / tiny_weights fixture against the new tape-authoritative path and an inline candle composite reference. CUDA-gated; skips cleanly when no CUDA device is visible. Asserts:
   - **Loss equality** (5e-4 tol): both paths run the same kt `cross_entropy` at the seed.
   - **The new path completes** without panicking or erroring out of the bridge.
   - **The returned GradStore is non-empty** (proves option-(a)'s `loss.backward()` populated the seed entry).

   **Not asserted**: per-LoRA-Var grad parity — see "Known gap" below.

## Known gap (handled by follow-up)

`compute_lora_delta` in `kiln-model::lora_loader` calls `.to_dtype(...)` + `.t()` + `.contiguous()` on the LoRA Vars before reaching the matmul adapter. Each call mints a fresh candle `TensorId`, so the matmul adapter registers intermediate ids — not the LoRA Var ids. The tape-authoritative path can't bridge the raw-id link back to the Vars without that coverage.

The candle-authoritative path papers over this by running candle backward, which walks back through the dtype/t/contiguous ops to populate Var grads. The tape-authoritative path doesn't have that fallback.

For now, the function warns loudly in this case (optimizer downstream will no-op for LoRA Vars). Wiring the LoRA-forward path through tape adapters that thread the Var ids directly is the next PR; that PR can also promote this test to a full per-Var grad-diff parity assertion (pattern in `test_checkpointed_reverse_gradients_match_standard_cpu`).

## Validation

- CPU workspace `cargo check` ✅
- CPU `cargo check -p kiln-train --tests` ✅
- A6000 `cargo check --release --features cuda -p kiln-train` ✅
- A6000 `cargo nextest run --release --features cuda -p kiln-train cuda_tape_authoritative_grads_match_standard_tiny`:
  ```
  PASS [   1.491s] (1/1) kiln-train trainer::tests::cuda_tape_authoritative_grads_match_standard_tiny
  ```

## Out of scope

- **LoRA-path tape adapter coverage**: the follow-up that closes the raw-id gap above.
- **FLCE wiring**: `try_tape_flce_phase_b_cuda` doesn't call `register_*` / `retain_output_for_chaining` yet. The new path errors with a clear message rather than silently failing inside the bridge.
- **GRPO / OPD / DPO**: this is SFT only. Each algorithm has its own loss-construction site that needs the same wiring.

## Refs

- #1082 (CP-4 endgame)
- `fb0c2f30`, `31208400` (Step B: `with_tape_authoritative_scope`)
- `fd61da18`, `d0c1d14d`, `3ca649c5` (Step A: kt-id chaining)
- Agent notes: `kiln-cp4-trainer-flip-turnkey`, `kiln-cp4-trainer-flip-gradstore-wrinkle`, `kiln-tape-bridge-seed-gap`, `kiln-tape-bridge-seeds-via-copy`

### DIFF
--- a/crates/kiln-autograd/src/lib.rs
+++ b/crates/kiln-autograd/src/lib.rs
@@ -97,4 +97,6 @@ pub use backwards::unsqueeze::UnsqueezeBackward;
 pub use backwards::where_select::WhereSelectBackward;
 pub use grad_store::GradStore;
 pub use tape::{Tape, TapeNode};
-pub use tape_scope::{tape_forward_enabled, with_active_tape, with_thread_local_tape};
+pub use tape_scope::{
+    tape_authoritative_enabled, tape_forward_enabled, with_active_tape, with_thread_local_tape,
+};
--- a/crates/kiln-autograd/src/tape_scope.rs
+++ b/crates/kiln-autograd/src/tape_scope.rs
@@ -84,6 +84,31 @@ pub fn tape_forward_enabled() -> bool {
     })
 }
 
+/// `KILN_TAPE_AUTHORITATIVE` env var — opt-in trainer flip (#1082 CP-4 Step 2).
+///
+/// When this is on, `kiln-train::trainer::standard_forward_backward` routes
+/// SFT backward through `kiln_kt_bridge::tape_bridge::with_tape_authoritative_scope`
+/// instead of candle's `loss.backward()`. The loss op (cross-entropy) must
+/// itself be tape-routed for the seed lookup to find a kt mirror — see the
+/// `try_tape_cross_entropy_cuda` adapter wired into
+/// `kiln-train::cross_entropy_loss`.
+///
+/// Implies `tape_forward_enabled` semantically: the authoritative path
+/// cannot work without tape adapters recording. The trainer's gate check
+/// asserts both before flipping. Off by default. Cached after first read
+/// (same `OnceLock` convention as [`tape_forward_enabled`]).
+pub fn tape_authoritative_enabled() -> bool {
+    static ENABLED: OnceLock<bool> = OnceLock::new();
+    *ENABLED.get_or_init(|| {
+        std::env::var("KILN_TAPE_AUTHORITATIVE")
+            .map(|v| {
+                let v = v.trim().to_lowercase();
+                !(v.is_empty() || v == "0" || v == "false" || v == "no")
+            })
+            .unwrap_or(false)
+    })
+}
+
 thread_local! {
     /// Thread-local active `Tape`. Wrapped in `RefCell` so the
     /// `record(...)` paths can take a mutable borrow; we hand out a
--- a/crates/kiln-train/src/trainer.rs
+++ b/crates/kiln-train/src/trainer.rs
@@ -5966,6 +5966,28 @@ fn cross_entropy_loss(
     // Cross-entropy: -log(softmax(logits)[label])
     // Use log-sum-exp trick for numerical stability
     let active_logits_f32 = active_logits.to_f32_dtype()?;
+
+    // CP-4 (#1082) Step 2: tape-authoritative wiring. When the tape forward
+    // gate is on AND a tape scope is active AND the inputs land in the
+    // adapter's CUDA envelope, route the loss through
+    // `try_tape_cross_entropy_cuda` so a `CrossEntropyKtBackward` node lands
+    // on the active `Tape`. `with_tape_authoritative_scope` then seeds the
+    // tape walk from this loss instead of running candle backward.
+    //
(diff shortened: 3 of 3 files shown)


ericflo:
Closing as fully superseded by the canonical landing in `43fe9c42` ("kiln-train: tape-authoritative SFT backward (KILN_USE_TAPE_AUTHORITATIVE) + parity gate (#1082 CP-4 endgame)"), which a parallel agent shipped at 23:54 UTC — about 10 minutes after I started exploring this task and before I pushed my branch. The canonical commit lands the same three deliverables (env-var flag, `standard_forward_backward_tape_authoritative` + GradStore option-(a), CE adapter wiring in `cross_entropy_loss`) and pairs cleanly with `62f43cb6` (`try_tape_lora_add_cuda` / `LoraDeltaAddBackward` / `MulScalarBackward`, #1412), which closes the LoRA-Var grad coverage gap my PR documented as follow-up.

Differences worth noting (informational only — canonical wins):

- The canonical flag is `KILN_USE_TAPE_AUTHORITATIVE` (fresh-read, no OnceLock cache), mine was `KILN_TAPE_AUTHORITATIVE` (OnceLock-cached, parallel to `tape_forward_enabled`). Fresh-read is the better choice for an opt-in trainer toggle — tests can flip it inline without process restart.
- The canonical CE-adapter gate is `tape_authoritative_enabled()` (so the candle-authoritative `with_tape_scope_emit_to_grad_store` path keeps a lineage-carrying candle loss); mine gated on `feature = "cuda"` alone, which would route the kt CE loss into the candle-authoritative path too — that path's grad-store iterator would not have matched it.
- The canonical parity gate asserts per-Var grad parity end-to-end thanks to `try_tape_lora_add_cuda`; mine softened to substrate-gate (loss equality + GradStore non-empty) because the LoRA-path raw-id link was missing at the time I tested.

Lesson logged for future agents: do a `git fetch origin main` + `gh pr list --search` ~halfway through implementation, not just at the start — same-day PRs can land while you're reading source. Saved as agent note `kiln-cp4-step2-concurrent-agent-collision` so future task plans flag the risk.

No follow-up needed from this PR.

PR #1435, opened 2026-06-01
ci: validate anomaly trap + deterministic envelope on training path

ericflo:
## What
Adds a Linux CI step that runs the per-PR SFT training smoke tests (`perf_regression_sft_train_*` — real forward + backward + optimizer on a tiny CPU fixture) with `KILN_DETECT_ANOMALY=1 KILN_DETERMINISTIC=1`.

## Why (#1082 boxes 128 / 495)
- Box 495: the `KILN_DETECT_ANOMALY` NaN/Inf trap is wired through the autograd tape (`kiln-autograd/src/tape.rs` backward loop + `anomaly.rs`) but was never exercised in CI. This turns it on in the training-parity CI lane, so a false-positive trap (or a regression that severs the trap) fails the build.
- Box 128: `KILN_DETECT_ANOMALY` and `KILN_DETERMINISTIC=1` are documented (ARCHITECTURE.md) but were not *validated in CI*. This step is that validation.

The smoke tests' existing asserts (adapter written, < 30 s) are the gate. Single-threaded so the process-global env vars don't race other tests. No GPU required — pure CPU training path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -96,6 +96,16 @@ jobs:
       - name: Test (default features)
         run: cargo test --locked -- --skip test_health_with_real_backend
 
+      - name: Test (anomaly trap + deterministic envelope on the training path — #1082 boxes 128/495)
+        # Validate KILN_DETECT_ANOMALY (the NaN/Inf trap wired through the
+        # autograd tape) and KILN_DETERMINISTIC=1 on a real CPU SFT step
+        # (forward + backward + optimizer). The perf-regression smoke tests'
+        # own asserts (adapter written, < 30 s) double as the gate: a
+        # false-positive anomaly trap or a broken deterministic codepath
+        # would fail the training step here. Single-threaded so the
+        # process-global env vars don't race other tests.
+        run: KILN_DETECT_ANOMALY=1 KILN_DETERMINISTIC=1 cargo test --locked -p kiln-train perf_regression_sft_train -- --test-threads=1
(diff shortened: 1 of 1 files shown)


ericflo:
**Blocked on the upstream default-build fix — not mergeable yet, but the change is correct.**

The `Linux / default features` and `macOS / Metal` jobs fail at `cargo build` with `error[E0433]: cannot find ... candle_core` in `kiln-model` — the in-progress candle removal currently breaks the default (no-feature) build on `main` (all recent main CI runs are red for the same reason). My new training-validation step runs *after* the build, so it never executes.

Net: this CI step (boxes #1082 128/495 — validate `KILN_DETECT_ANOMALY` + `KILN_DETERMINISTIC` on the CPU SFT path) cannot go green until `kiln-model` compiles on default features again (i.e. once the candle-removal finale lands). The step itself is correct and self-contained; leaving this PR open so it validates + merges the moment the default build is restored. Not checking boxes 128/495 until this actually runs green.

ericflo:
Superseded by #1438. This PR validated the anomaly trap + deterministic envelope via `cargo test -p kiln-train`, but kiln-train pulls kiln-model — which can't compile mid-candle-migration — so the `linux-default` job fails before this step runs. #1438 validates the same env vars on the **candle-free** crates (kiln-autograd anomaly probe + kiln-tensor determinism selector) where they actually live, so the check executes today on a GPU-less runner.

PR #1460, opened 2026-06-04
serve: default CUDA graphs ON for bs=1 decode (#34)

ericflo:
The payoff for the #34 BUG2 work: **CUDA graphs default-ON** for single-sequence decode (~10–15% decode speedup), no env opt-in needed.

Both bs=1 blockers are fixed:
- **BUG1** (dangling captured block-table → `CUDA_ERROR_ILLEGAL_ADDRESS`) — fixed earlier by the default-on stable paged-metadata buffer.
- **BUG2** (token-doubling on replay) — fixed in #1457 by instantiating with `flags=0` instead of `AUTO_FREE_ON_LAUNCH` (which re-allocated the GDN recurrent/conv state buffers each replay, desyncing the in-place state writes). Root-caused by diffing against the **working ROCm HIP-graph path** (which uses `flags=0` and is bit-identical to eager), and confirmed to compile in a full `kiln-model --features cuda` build on a fresh A6000.

`MemoryConfig::default().cuda_graphs` flips `false → true` (matching its long-standing doc comment "Default: true"). **Escape hatch:** `KILN_CUDA_GRAPHS=0`. The bs>1 batched path stays opt-in (`KILN_CUDA_GRAPHS_BATCHED`).

**Honest caveat:** the on-CUDA *capture-vs-eager bit-identical decode* runtime check still wants a run with the model loaded (needs the 9 GB custom model on a CUDA box). The fix is root-caused, compiles on real CUDA, and mirrors the proven ROCm fix; the env flag is the safety valve if anything regresses. `test_defaults` updated; explicit-false TOML + env-override tests unchanged (26/0 pass).

### DIFF
--- a/crates/kiln-server/src/config.rs
+++ b/crates/kiln-server/src/config.rs
@@ -485,8 +485,13 @@ impl Default for MemoryConfig {
             inference_memory_fraction: 0.7,
             training_memory_gb: None,
             kv_cache_fp8: false,
-            // Opt in until graph-capturable stream replay matches eager decode.
-            cuda_graphs: false,
+            // Default ON (#34): both bs=1 graph bugs are fixed — BUG1 (dangling
+            // captured block-table) by the default-on stable paged metadata, and
+            // BUG2 (token-doubling on replay) by instantiating the graph with
+            // flags=0 instead of AUTO_FREE_ON_LAUNCH (which re-allocated the GDN
+            // state buffers each replay). Disable with `KILN_CUDA_GRAPHS=0`. The
+            // bs>1 batched path stays opt-in (`KILN_CUDA_GRAPHS_BATCHED`).
+            cuda_graphs: true,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


ericflo:
HOLD — do NOT merge. A deeper investigation surfaced that this default-ON is premature: `bench-results/cuda-graph-box102-findings.md` (runtime-instrumented on A6000) explicitly RULED OUT the `flags=0` change as the BUG2 fix (finding #3: flags=0 produced identical token-doubling). The real root cause is a stale `token_buffer`/`position_buffer` read by the captured embedding on replay (layer-0 divergence; replay logits differ ~4.4% from eager). The bs=1 graph path still produces token-doubling garbage on CUDA, so it must NOT be default-ON. Converting to draft until the real bug is fixed AND runtime-verified bit-identical on hardware.

ericflo:
CLOSING. Runtime test on A6000 (Qwen3.5-4B, BF16 + W4A16, KILN_BATCHING_ENGINE=0, 256-token decode) shows CUDA graph bs=1 decode is NOT bit-identical to eager — it diverges from eager on 2/3 varied prompts with BOTH flags=0 AND AUTO_FREE_ON_LAUNCH (a ~4.4% replay-logit difference flips an argmax mid-stream → different-but-coherent output). box102 was right; the real bug (layer-0/input replay staleness) is unsolved and deeper than the instantiate flag. Default-ON is unsafe until that's fixed and verified bit-identical.

PR #1461, opened 2026-06-04
Revert cuda-graph flags=0 "BUG2 fix" — disproven by box102 (#34)

ericflo:
My #34 `flags=0` change (in #1457) was **wrong** and had already been disproven: `cuda-graph-box102-findings.md` finding #3 tested `flags=0` on A6000 and saw **identical token-doubling**. It's orthogonal to BUG2 (the real bug is layer-0/input replay staleness). Restore `AUTO_FREE_ON_LAUNCH` at both sites — it correctly reclaims per-launch scratch, and on the batched path flags=0 would leak it. CUDA graphs stay default-OFF; #1460 (default-ON) is held as draft until the real bug is fixed + runtime-verified.

### DIFF
--- a/crates/kiln-model/src/cuda_graph.rs
+++ b/crates/kiln-model/src/cuda_graph.rs
@@ -1889,23 +1889,21 @@ impl CudaGraphRunner {
             })
         });
 
-        // End capture — instantiates the graph.
+        // End capture — instantiates the graph with AUTO_FREE_ON_LAUNCH.
         //
-        // (BUG2 fix) Instantiate with NO flags (0), NOT AUTO_FREE_ON_LAUNCH.
-        // AUTO_FREE_ON_LAUNCH frees the graph's per-launch async allocations at
-        // the start of every replay and re-allocates them — so the GDN
-        // recurrent/conv-state kernels write their new state into freshly-
-        // allocated scratch that the persistent recurrent_states[i]/conv_states[i]
-        // device pointer no longer aliases. The state therefore never advances
-        // across bare graph.launch() replays, freezing the decode and re-emitting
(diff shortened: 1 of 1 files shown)


ericflo:
HOLD — contested. The cuda-vs-rocm-replay diff (using the WORKING ROCm path as oracle) argues this revert is wrong: ROCm instantiates with NO flags and is bit-identical to eager, making AUTO_FREE_ON_LAUNCH the ONLY graph-level difference. box102's flags=0 disproof was contaminated — it was run on wip/1082-box2-fix-validate WITH the slice_set recurrent-state wrapper + FORCE_ZERO/SAMESTEP probes active, which mutate the GDN state being measured. The real fix is likely flags=0 (keep) + default-on per-replay arena memset (Fix B). Settling this with a CLEAN A6000 runtime test before merging either direction.

ericflo:
CONFIRMED ON HARDWARE — un-drafting + merging. The A6000 runtime test settles the flags=0-vs-AUTO_FREE question definitively: a 256-token greedy decode (Qwen3.5-4B, BF16 + W4A16, KILN_BATCHING_ENGINE=0) diverges from eager on 2/3 prompts with BOTH flags=0 AND AUTO_FREE_ON_LAUNCH. So flags=0 is NOT the BUG2 fix (box102 was right; the diff-workflow's 'box102 contaminated' theory is disproven by runtime). Since both are equally (in)correct and AUTO_FREE is the original deliberate choice that reclaims per-launch scratch, restoring it is correct. The real bug — a ~4.4% replay-logit divergence (layer-0/input staleness) — remains unsolved; CUDA graphs stay default-OFF.

PR #1474, opened 2026-06-10
ci: run all 163 kiln-vulkan-kernel tests in the Vulkan job (was: 18)

ericflo:
## Problem

The **Linux / Vulkan** job's test step was restricted to `--test gdn_parity` — 18 of the crate's 163 tests. Everything else (`vk_matmul_parity`, `vk_attention_parity`, `vk_flce_parity`, `vk_opd_parity`, `vk_gdn_backward_parity`, …) never executed in CI. Most notably, `vk_muon_parity` — the parity tests for the fused kernel of the **new default optimizer** (#1468) — has zero CI coverage.

## Change

Drop the `--test gdn_parity` restriction so the lib unit tests + all 20 integration targets run. The job already installs `mesa-vulkan-drivers` (lavapipe) and executes the gdn_parity kernels for real (~0.5s in run 27294478586), so the rest run under the same software device.

## Validation

- Local Strix Halo (RADV): `cargo test --locked -p kiln-vulkan-kernel` → **163 passed, 0 failed**, all targets <0.1s each.
- This PR's own Linux / Vulkan job is the lavapipe proof. If a specific test trips on a lavapipe quirk (e.g. subgroup-size assumptions), the fallback is a targeted `--skip` with a comment — not re-restricting the suite.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,7 +205,12 @@ jobs:
         run: cargo check --locked -p kiln-server --features vulkan
 
       - name: Test Vulkan kernel crate
-        run: cargo test --locked -p kiln-vulkan-kernel --test gdn_parity -- --nocapture
+        # Full suite (lib unit tests + every tests/ target), not just
+        # gdn_parity — this is the only CI leg that executes the Vulkan
+        # kernels (under lavapipe), and the restriction left e.g. the
+        # vk_muon_parity tests for the default optimizer's fused kernel
+        # never running in CI. 163 tests, <10s total.
+        run: cargo test --locked -p kiln-vulkan-kernel -- --nocapture
 
   linux-rocm:
     name: Linux / ROCm (build)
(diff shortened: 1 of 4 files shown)


ericflo:
The Linux/Vulkan failure was the anticipated lavapipe quirk, but the root cause is a real contract gap rather than a flaky test: `VkPagedKvCache::new` documents *"Allocate a zero-initialized cache"* but never zeroes the pools — the property held only because RADV hands back kernel-zeroed pages. lavapipe hands back malloc garbage, so `paged_kv_write_slot_resident_writes_one_slot_exactly` (which asserts non-target slots stay 0.0) failed.

Pushed a fix making the documented contract real: `VulkanBuffer::fill_zero` (one-shot `vkCmdFillBuffer`, no staging) called for every K/V layer pool at allocation. 163/163 locally on RADV; this CI run validates lavapipe.

